Issue #7, opened 2021-09-01
Relicense fully under Apache-2.0 WITH LLVM-exception


cfallin:
Currently, regalloc2 contains some MPL code. This code was a result of an initial translation effort that started from IonMonkey's register allocator; IonMonkey (as part of SpiderMonkey and thus the Firefox codebase) is licensed under the MPL. This is a perfectly fine open-source license, but ideally, we would like to relicense this code to fall under `Apache-2.0 WITH LLVM-exception`, in alignment with other BA codebases, if all of the relevant parties agree and are happy with this.

cc @tschneidereit, who has been in contact with various folks about this and is working hard to obtain the appropriate signoffs (thanks!).

PR #9, opened 2021-09-03
Remove Function::is_call

Amanieu:
The documentation says that this is only used for heuristics, but it
is never actually called. This should be removed for now and perhaps
added back later if we find an actual use for it.

### DIFF
--- a/doc/DESIGN.md
+++ b/doc/DESIGN.md
@@ -49,9 +49,8 @@ successors.
 
 Instructions are opaque to the allocator except for a few important
 bits: (1) `is_ret` (is a return instruction); (2) `is_branch` (is a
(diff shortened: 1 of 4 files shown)


cfallin (approved):
Thanks! `is_call` was left over from an earlier point when calls were considered as liverange split-points; the "split at first conflict" heuristic is much simpler though and obsoleted the need to know about calls.

cfallin:
The `cargo-deny` CI errors have to do with licenses and I think we can look at this more closely when #7 is complete; until then, I'll do a PR to remove the check and then we can rebase this over it.

cfallin:
I created #10 for the cargo-deny issue.

PR #11, opened 2021-09-05
Simplify Requirement by removing register classes

Amanieu:
We never merge bundles from vregs of different classes, so we don't
need to check for register class conflicts.


cfallin (approved):
LGTM, thanks! Happy to merge once this is rebased to include the cargo-deny fix.

Issue #12, opened 2021-09-05
Support explicit stack slots


Amanieu:
This is useful for functions which take many arguments and return many values, some of which are in registers while the rest are in ABI-specified stack slots.

I think the best way to add support for fixed stack slots is to add an extra bit to `PReg` to add 32 integer stack slots and 32 float stack slots.

Implementation-wise, a fixed stack slot is very similar to `Operand::Fixed`:
- The same fixups are required in case of multiple `Fixed` uses of the same vreg.
- Live ranges assigned to a fixed stack slot need to tracked just like a `PReg`.
- Allocation into fixed stack slots must be done in `process_bundle`: we may need to split a bundle in case of conflicts or evict another conflicting bundle.
- Requirement computation only needs minor changes to merge `Requirement::Stack` with `Requirement::Fixed(preg)` when `preg` refers to a fixed stack slot.
- A spill weight of 2000 still makes sense for fixed stack slots: we need expensive moves if we don't get the stack slot that we want.

cfallin:
Sorry for the delay in thinking about this -- it's a worthwhile goal, I think! I'm definitely onboard with some sort of interface that allows the user to name explicit stack slot locations.

It seems to me that this scheme is, more or less, defining virtual registers (in the general sense of "virtual", not specifically "VRegs", if that makes sense): we basically have a new class, or register file, and we allocate from its entries/slots like any other. Then these are used in the same way other ABI-specified registers are used for call args. Is that accurate?

I had worried that supporting a more general explicitly-specified-stack-slot need would require a larger index (in general `SpillSlot` needs to grow up to the number of live values, which can be large) but on more reflection I can see that your scheme seems to cover most cases where the user cares about which spillslot specifically is used.

In Cranelift at least this will a really nice way to avoid the explicit load of stack-passed args in the prologue. Presumably your use-case is similar. (If there are more than 32 stack args of either class, we can still fall back on generally unconstrained vregs and explicit loads in the prologue.)

I'm unsure whether it makes more sense to define actual `RegClass` values (`IntStack` and `FloatStack`) or make this another dimension of the `PReg`, but I'm sort of leaning toward the latter, as it sounds like you are, as well.

[...]

Issue #5, opened 2021-09-01
Investigate better bitpacking for Operand and Use


cfallin:
Two core data-structure elements, `Operand` and `Use`, are both designed to fit a relatively large amount of information in one `u32`. This is a performance optimization that we have found to be relatively impactful; expanding even to a `u64` has a measurable impact (of at least a few percent) on compilation time.

Unfortunately, the scarcity of bits means that certain limits are lower than we would prefer. For example, we support only a 5-bit index for physical registers in each register class (so 32 integer registers and 32 float/vector registers), which may not be enough for some use-cases (though it can work for aarch64 and x64 at least). This also limits the VReg count to 1M (2^20).

We should investigate ways of, e.g., out-of-lining infrequently-used information (such as fixed-PReg constraints) to raise the limits on VRegs, PRegs, instruction count, and the like and provide enough headroom for any reasonably-imaginable use case.

Amanieu:
When we remove non-SSA support (#4), we can shrink `OperandKind` to 1 bit instead of 2. Also, we can change the encoding of `constraint` + `preg` to take only 6 bits instead of 8:
- `1xxxxx` => `Fixed(preg)`
- `01xxxx` => `Reuse(index)`
- `000000` => `Any`
- `000001` => `Reg`
- `000010` => `Stack`
- `_` => Free for future use

This bumps the limit for the VReg index to 8M (2^23). However I don't see how we can fit explicit stack slots into this scheme.

To go even further, we would need to store auxillary `Operand` information in a separate `u8`. By moving `preg`/`index` into the second word, we would leave 26 bits for the VReg index which allows for up to 64M VRegs (the remaining 6 bits are: `kind`, `pos`, `class` and `constraint:3`). The size of `Use` wouldn't increase since we have a padding byte to spare. This also allows us to have up to 256 PRegs per class instead of 32, 256 explicit stack slots, etc.

```rust
struct OperandExtra(u8);

trait Function {
    // Both slices must have the same length.
    fn inst_operands(&self, insn: Inst) -> (&[Operand], &[OperandExtra]);
}
```

cfallin:
Allowing a second `u32`, even, tagged with an instruction index, seems like it could be a reasonable approach to allow more information to be provided in "rare" cases (stackslots, fixed pregs). I guess at this point we're really designing a variable-length compression scheme so we should be somewhat data-driven on that question, of course.

[...]

PR #13, opened 2021-09-13
Adjust Operand encoding

Amanieu:
The encoding for OperandConstraint is adjusted to free up 2 bits which
allows for 2^21 vregs and 2^6 pregs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,8 +60,8 @@ pub enum RegClass {
 /// The value returned by `index()`, in contrast, is in a single index
 /// space shared by all classes, in order to enable uniform reasoning
 /// about physical registers. This is done by putting the class bit at
-/// the MSB, or equivalently, declaring that indices 0..63 are the 64
-/// integer registers and indices 64..127 are the 64 float registers.
+/// the MSB, or equivalently, declaring that indices 0..=63 are the 64
+/// integer registers and indices 64..=127 are the 64 float registers.
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct PReg {
     hw_enc: u8,
@@ -383,7 +383,6 @@ pub enum OperandPos {
 /// that the conflict (overlap) is properly accounted for. See
 /// comments on the constructors below for more.
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-//#[repr(packed)]
 pub struct Operand {
     /// Bit-pack into 32 bits.
     ///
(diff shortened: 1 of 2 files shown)


Amanieu:
Extending the number of pregs is a prerequisite for #12 and increasing the number of vregs helps a bit towards #5.

src/lib.rs
```diff
@@ -55,21 +55,21 @@ pub enum RegClass {
...
 /// Because of bit-packed encodings throughout the implementation,
-/// `hw_enc` must fit in 5 bits, i.e., at most 32 registers per class.
+/// `hw_enc` must fit in 6 bits, i.e., at most 64 registers per class.
 ///
 /// The value returned by `index()`, in contrast, is in a single index
 /// space shared by all classes, in order to enable uniform reasoning
 /// about physical registers. This is done by putting the class bit at
-/// the MSB, or equivalently, declaring that indices 0..31 are the 32
-/// integer registers and indices 32..63 are the 32 float registers.
+/// the MSB, or equivalently, declaring that indices 0..63 are the 64
```
cfallin, line 63: Pre-existing but a pedantic nit on my own comment: could you adjust to either `0..=63` or `0..64` (and likewise for the second range)? Inconsistency w.r.t. Rust open/cosed-range syntax is bothering me :-)

src/lib.rs
```diff
@@ -404,29 +404,28 @@ impl Operand {
         kind: OperandKind,
         pos: OperandPos,
     ) -> Self {
-        let (preg_field, constraint_field): (u32, u32) = match constraint {
-            OperandConstraint::Any => (0, 0),
-            OperandConstraint::Reg => (0, 1),
-            OperandConstraint::Stack => (0, 2),
+        let constraint_field = match constraint {
```
cfallin, line 407: Could you add a block comment here describing the bit encoding?

cfallin (approved):
Thanks for this!

Just a request for an additional comment below but otherwise LGTM.

src/lib.rs
```diff
@@ -383,15 +383,15 @@ pub enum OperandPos {
 /// that the conflict (overlap) is properly accounted for. See
 /// comments on the constructors below for more.
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+//#[repr(packed)]
```
lqd, line 386:
```suggestion
```

Amanieu: Oops that's from an old test I was doing.

cfallin (approved):
LGTM!

PR #16, opened 2021-11-13
Various fixes & minor improvements

Amanieu:
- Fixed a bug from #13 with PRegs with index > 32
- Removed `regs` from `MachineEnv` since it's not actually needed.
- Return safepoint_slots as Allocations instead of SpillSlots .
  - This enables us to support reftype vregs in register locations in the
future.
-  Removed an unused `clobbers` vector.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -78,7 +78,7 @@
 
 use crate::{
     Allocation, AllocationKind, Block, Edit, Function, Inst, InstPosition, Operand,
-    OperandConstraint, OperandKind, OperandPos, Output, PReg, ProgPoint, SpillSlot, VReg,
+    OperandConstraint, OperandKind, OperandPos, Output, PReg, ProgPoint, RegClass, VReg,
 };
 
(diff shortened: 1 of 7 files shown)


cfallin (approved):
This all looks reasonable, thanks! The generalization to reftypes-in-`Allocation` is good and will eventually be useful (for use-cases where the runtime supports it). IIUC, this hasn't changed where refs are actually placed at safepoints though, right? It might be worth adding a note in the `safepoint_slots` doc-comment noting that for now, all values will be in stackslots, until a future option is added (per-safepoint or for the whole function).

With that change, 👍 from me!

Also, to ensure complete explicitness and clarity, re: the relicensing that is ongoing in #7 currently, you are fine with this contribution being under the new license as well, yes?

Thanks!

Amanieu:
I've updated the doc comment for `safepoint_slots`.

And yes, I'm fine with relicensing this as described in #7.

cfallin:
Thanks!

PR #17, opened 2021-11-28
Add support for fixed stack slots

### LINKED ISSUE #12: Support explicit stack slots

Amanieu:
This PR adds the ability to mark a `PReg` as being a fixed stack slot through `MachineEnv`. This means that:
- The `PReg` can be specified using `OperandConstraint::FixedReg`.
- This `PReg` can be selected for `OperandConstraint::Stack`, but not `OperandConstraint::Reg`. It also cannot be used with `OperandConstraint::Reuse`, which only works with a `OperandConstraint::Reg`.
- Fixed stack slot allocations are returns as an `Allocation::reg`, just like a normal fixed stack slot.
- A `VReg` marked as `requires_refs_on_stack` can be placed into a fixed stack slot instead of a spillslot.

I'm not too happy about the API (`PReg` stands for "physical register", which feels wrong here) but it was the cleanest way to integrate this into the rest of the codebase.

Fixes #12


src/ion/liveranges.rs
```diff
@@ -1159,66 +1180,108 @@ impl<'a, F: Function> Env<'a, F> {
...
+                            // Reg/FixedReg constraints. FixedReg is
+                            // incompatible if there already is a different
+                            // FixedReg constraint.
+                            if !(requires_reg && self.pregs[preg.index()].is_stack)
+                                && *first_preg.get_or_insert(preg) == preg
+                            {
+                                continue;
+                            }
+
+                            log::trace!(" -> duplicate; switching to constraint Reg");
```
cfallin, line 1264: I think it might be pre-existing but the trace message should say we're switching to `Any`, no?

src/ion/liveranges.rs
```diff
@@ -1159,66 +1180,108 @@ impl<'a, F: Function> Env<'a, F> {
...
+                                "at pos {:?}, vreg {:?} has fixed constraint to preg {:?}",
+                                u.pos,
+                                vreg_idx,
+                                preg_idx
+                            );
+
+                            // FixedStack is incompatible if there are any
+                            // Reg/FixedReg constraints. FixedReg is
+                            // incompatible if there already is a different
+                            // FixedReg constraint.
```
cfallin, line 1257: A little more to this comment would help understanding I think: "If either condition is true, we edit the constraint below; otherwise, we can skip this edit."

src/ion/requirement.rs
```diff
@@ -31,39 +32,53 @@ impl Requirement {
...
             OperandConstraint::Reg | OperandConstraint::Reuse(_) => Requirement::Register,
             OperandConstraint::Stack => Requirement::Stack,
             _ => Requirement::Any,
         }
     }
-}
 
-impl<'a, F: Function> Env<'a, F> {
-    pub fn compute_requirement(&self, bundle: LiveBundleIndex) -> Requirement {
+    pub fn compute_requirement(&self, bundle: LiveBundleIndex) -> (Requirement, ProgPoint) {
```
cfallin, line 65: A doc-comment describing the return value here would be useful. Actually, a struct might be nice: something like

```
struct RequirementResult {
    pub req: Requirement,
    pub needs_split: Option<ProgPoint>,
}
```

for readability maybe?

cfallin: (Good idea to return the splitpoint eagerly rather than re-scanning to find it later, by the way -- every little efficiency improvement counts!)

src/ion/requirement.rs
```diff
@@ -31,39 +32,53 @@ impl Requirement {
...
                 req = req.merge(r);
                 log::trace!("     -> req {:?}", req);
+                if req == Requirement::Conflict {
+                    return (req, u.pos);
+                }
             }
         }
         log::trace!(" -> final: {:?}", req);
-        req
+        (req, ranges.first().unwrap().range.from)
```
cfallin, line 82: Perhaps I slightly misunderstand the purpose of the `ProgPoint`, given this -- in this case we don't need a split, so I guess we can return `None` (if `Option` as I suggested above)?

src/ion/liveranges.rs
```diff
@@ -1159,66 +1180,108 @@ impl<'a, F: Function> Env<'a, F> {
...
-                    usize,
-                )>| {
-                    if last_point.is_some() && Some(pos) != last_point {
-                        seen_fixed_for_vreg.clear();
-                        first_preg.clear();
+
+                // Find groups of uses that occur in at the same program point.
+                for uses in self.ranges[range.index()]
+                    .uses
+                    .group_by_mut_(|a, b| a.pos == b.pos)
```
cfallin, line 1187: This is a nice API but it's unfortunate that it's not yet stable, and that we have to copy part of stdlib into `util.rs` to get it; I'd prefer not to duplicate like that, and to add a few hundred lines of code to save just a few here. Until we have stabilized `group_by_mut`, could we just do the scan manually here?

cfallin (commented):
Thanks so much for this contribution!

I have a few minor comments, but this looks great overall, very carefully done.

src/ion/liveranges.rs
Amanieu: I actually tried to do the scan manually at first and the result was an unreadable mess. Hiding the logic to group entries by `ProgPoint` makes the loop *much* more readable.

I can shrink down the code in `util.rs` to only contain the parts that we actually needs (e.g. we don't use `group_by`, only `group_by_mut`). Alternatively, the itertools crate also provides a version of `group_by` (with slightly different arguments).

cfallin: Ah, I think using `itertools` is reasonable for now, then; it's a pretty small dep and if it improves readability here, then all the better. (IIRC there are similar scan-and-find-groups patterns in e.g. move resolution; maybe we can clean those up too later, but that's obviously outside the scope of this PR!)

Amanieu:
I've applied the feedback from the review and also done a bit more cleanup on the API for interacting with requirements (it uses `Result::Err` to report a conflict, which is more Rust-y).

cfallin (approved):
Looks great, thanks!

Issue #8, opened 2021-09-01
Rematerialization


Amanieu:
As an alternative to spilling and reloading, a vreg value can sometimes be recomputed from other available values. There are two ways this can be supported in regalloc2:

## Simple rematerialization

Supporting rematerialization for constants, including runtime constants which only depend on pinned registers (e.g. VMContext), is relatively straightforward. The client needs to track rematerialization metadata for each `VReg`, after which we can elide spills for this vreg (unless there is a stack use) and replace reloads with rematerializations.

```rust
trait Function {
    fn can_rematerialize(&self, vreg: VReg) -> bool;
}

enum Edit {
    /// The code sequence for rematerialization is only allowed to
    /// use `dest` and the dedicated scratch register.
    Rematerialize {
        vreg: VReg,
        // This is not an Allocation because rematerializing
        // into a stack slot doesn't make sense.
        dest: PReg,
    }
}
```

## Complex rematerialization

Rematerialization involving values in other vregs is more complex and probably not worth the effort of implementing. It has been done before as part of a [research project on V8](https://csit.am/2015/proceedings/PDC/PDC5.pdf) but AFAIK this has not made it into the main V8 codebase.

Amanieu:
I'm currently looking to implement the simple rematerialization scheme I described above. Here's a summary of the changes I think are needed:
- In `compute_liveness`, don't add uses to a liverange for block parameters on branches. These aren't real operands so it's fine if the allocator assigns `Allocation::none` to them. This seems to already be the case for moves. This is needed when a vreg has no current location but will be rematerialized later.
- In `spill_weight_from_constraint`, assign a spill weight of 0 to defs of rematerializable vregs and half of the usual value for uses.

- `LiveBundle` needs a "remat" flag if it is associated with a rematerializable vreg. This flag is cleared if the bundle is merged with another vreg or if the bundle has any stack uses.
- In `process_bundle`, treat `Requirement::Any` as `Requirement::Register` for "remat" bundles. We prefer allocating `Any` constraints to registers when a stack slot is not required.
- In `try_allocating_regs_for_spilled_bundles`, don't mark a spillset as required for spilled bundles that have the "remat" flag.

[...]

PR #18, opened 2021-11-30
Rework the API for outgoing blockparams

Amanieu:
Branch parameters are not proper operands, so it doesn't make sense to represent them using `Operand`s. The constraint is irrelevant (you always want `Any`), and so is the position within the instruction (early/late).

This PR adds `Function::branch_params` which returns the branch parameters directly as a list of `VReg`s.

This is also useful for rematerialization (#8): branch parameters don't need to be in an actual register or stack slot if we can rematerialize them later. This would normally be required by the `Any` constraint used for branch parameters.

### DIFF
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -101,7 +101,7 @@ impl CFGInfo {
             }
             if require_no_branch_args {
                 let last = f.block_insns(block).last();
-                if f.branch_blockparam_arg_offset(block, last) > 0 {
+                if !f.inst_operands(last).is_empty() {
                     return Err(RegAllocError::DisallowedBranchArg(last));
                 }
             }
--- a/src/fuzzing/func.rs
+++ b/src/fuzzing/func.rs
@@ -13,7 +13,6 @@ use arbitrary::{Arbitrary, Unstructured};
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum InstOpcode {
-    Phi,
     Op,
     Ret,
     Branch,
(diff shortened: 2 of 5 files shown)


cfallin (approved):
Thanks for this! I was unsure at first about this refactor (the appeal of the blockparams approach to me is that phis turn into normal operands at their points of logical use, but... now they aren't operands?) but I think this is actually the right abstraction, on further thought: blockparams are indeed differently-constrained (they aren't really a use that must read the value, they just plumb it through), and the scheme for packing them contiguously in the args of a branch instruction was awkward. And blockparam inputs are special, so why not blockparam outputs too?

IIRC, an earlier iteration of RA2 actually depended on the uses existing, but the current implementation of blockparam plumbing makes use of the unified move generation so I think everything should work as it is now. Just to check: did you run the fuzzer for a bit with the updated logic?

Amanieu:
I did run the fuzzer, but looking at the code again I think that the live ranges are not being tracked properly. I need to double-check this.

cfallin:
Ah, yes, the dataflow analysis around `liveranges.rs` line 333 needs to consider out-blockparams as well, I think.

Amanieu:
Fixed, but I'll leave the fuzzer running overnight again to be sure.

Amanieu:
Fuzz tested ~4M test cases and no problems found! But then again fuzzing was also passing without the liverange fixes...

cfallin:
Sounds good. I'm not sure why the fuzzer didn't catch that -- perhaps it's likely to cover liveranges anyway by randomly using lots of values from earlier in the program... in any case, the logic looks correct now.

PR #21, opened 2021-12-05
Fix fuzzbug: add checker metadata for new vreg on multi-fixed-reg fixup move.

cfallin:
When an instruction uses the same vreg constrained to multiple different
fixed registers, the allocator converts all but one of the fixed
constraints to `Any` and then records a special fixup move that copies
the value to the other fixed registers just before the instruction. This
allows the allocator to maintain the invariant that a value lives in
only one place at a time throughout most of its logic, and constrains
the complexity-fallout of this corner case to just a special last-minute
edit.

Unfortunately some recent CPU time thrown at the fuzzer has uncovered
a subtle interaction with the redundant move eliminator that confuses
the checker.

Specifically, when the correct value is *already* in the second
constrained fixed reg, because of an unrelated other move (e.g. because
of a blockparam or other vreg moved from the original), the redundant
move eliminator can delete the fixup move without telling the checker
that it has done so.

Such an optimization is perfectly valid, and the generated code is
correct; but the checker thinks that some other vreg (the one that was
copied from the original) is in the second preg, and panics.

The fix is to use the mechanism that indicates "this move defines a new
vreg" (emitting a `defalloc` checker-instruction) to force the checker
to understand that after the fixup move, the given preg actually
contains the appropriate vreg.

<details>
<summary>Fuzzbug input (base64-encoded)</summary>
```
AAAAAwVrAAAAAAAAAAAAAFdXV1dXV1dXV1dXV1dXV1dXV1dXV1dXV1dXVxdXV1cBCldXV1dfV1dX
V1cAV1dXV1dXV1dXVwBXV0xNTVdXV1dXV1dXV1dXV1dXV7lXV1dXV1cAAAAAAAAAACQAAAAAAAAg
AAgAACkAAAAAAAAAAPoAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADPAAAAAAAAAAAAAAAAAAAAACQA
AAAAAAAgAAgAACkAAAAAAAD/////////////////////////////////////////////////////
/////////////////////////////////////////////wAA+gAAAAAAAAAAAAAAAKb9/////wD/
//+G/////1sL/10Ae90AAAAAAAAgAAAAAAAAYAAAAAAAAAAApaWlpaWlpaWlpaWlpaWlpaWlpaWl
pQClpaWlpaWlpaWlAA0NDQ0NDQ0NDf///1dXV1dXV1dXV1dXV1dXV1dXV7lXV1dXV1cAAAAAAAAA
ACQAAAAAAAAgAAgAACkAAAAAAAAAAPoAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAACwAA
AAD/XQB7
```
</details>

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -331,7 +331,7 @@ pub struct Env<'a, F: Function> {
     // was to the approprate PReg.
     //
     // (progpoint, copy-from-preg, copy-to-preg, to-slot)
-    pub multi_fixed_reg_fixups: Vec<(ProgPoint, PRegIndex, PRegIndex, usize)>,
+    pub multi_fixed_reg_fixups: Vec<(ProgPoint, PRegIndex, PRegIndex, VRegIndex, usize)>,
 
     pub inserted_moves: Vec<InsertedMove>,
 
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -1167,6 +1167,7 @@ impl<'a, F: Function> Env<'a, F> {
                     ProgPoint,
                     PRegIndex,
                     PRegIndex,
+                    VRegIndex,
                     usize,
                 )>| {
                     if last_point.is_some() && Some(pos) != last_point {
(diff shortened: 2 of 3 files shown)


cfallin:
@Amanieu would you mind reviewing? Thanks! (I'm also throwing some CPU at the fuzzer for a bit, inspired by your recent comments on doing so)

Amanieu (approved):

PR #23, opened 2021-12-27
Add a helper to iterate over insts and edits of a block in order

Amanieu:
This adds a `block_insts_and_edits` function on `Output` which produces an iterator over the instructions and edits of a block after register allocation. It's a very useful helper function for any consumers of the regalloc output.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -77,8 +77,8 @@
 #![allow(dead_code)]
 
 use crate::{
-    Allocation, AllocationKind, Block, Edit, Function, Inst, InstPosition, Operand,
-    OperandConstraint, OperandKind, OperandPos, Output, PReg, ProgPoint, RegClass, VReg,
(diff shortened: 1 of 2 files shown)


cfallin (approved):
This is quite useful, thanks!

I wonder if it would make sense to store the index into the edit-list corresponding to the start of each block (we could easily build this during the scan that produces all edits); that would eliminate the need to do the binary search. I don't feel too strongly about this though and so I'm fine with this as-is if you prefer it this way.

Amanieu:
We don't generate the moves in order: blockparam moves are inserted at the end, and then we sort the vector of moves by progpoint. We would need to perform the same kind of binary search to build the offsets into the edit list for each block.

I think the current solution works well enough if the edit list is only iterated once for each block.

cfallin:
Ah, yes, you're right -- this is good enough then, I think. Thanks!

PR #25, opened 2022-01-11
Performance improvements

Amanieu:
Overall these changes improve the performance of the register allocator by about 20%. See the individual commits for details.

### Before
```
         13,499.51 msec task-clock                #    1.000 CPUs utilized          
                50      context-switches          #    3.704 /sec                   
                15      cpu-migrations            #    1.111 /sec                   
            29,927      page-faults               #    2.217 K/sec                  
    57,567,174,127      cycles                    #    4.264 GHz                    
   132,885,501,992      instructions              #    2.31  insn per cycle         
    25,862,250,921      branches                  #    1.916 G/sec                  
       606,424,261      branch-misses             #    2.34% of all branches        
```

### After
```
         12,692.32 msec task-clock                #    1.000 CPUs utilized          
                55      context-switches          #    4.333 /sec                   
                 7      cpu-migrations            #    0.552 /sec                   
            29,118      page-faults               #    2.294 K/sec                  
    49,416,356,199      cycles                    #    3.893 GHz                    
   108,527,679,813      instructions              #    2.20  insn per cycle         
    20,600,616,983      branches                  #    1.623 G/sec                  
       574,496,164      branch-misses             #    2.79% of all branches 
```


fuzz/Cargo.toml
```diff
@@ -49,3 +49,9 @@ name = "ion_checker"
 path = "fuzz_targets/ion_checker.rs"
 test = false
 doc = false
+
+# Enable debug assertions and overflow checks when fuzzing
+[profile.release]
+debug = true
+debug-assertions = true
+overflow-checks = true
```
bjorn3, line 57: nit: missing trailing newline

Amanieu:
Added another optimization, speedup is now ~20%.

src/lib.rs
```diff
@@ -12,6 +12,16 @@
 
 #![allow(dead_code)]
 
+// Even when trace logging is disabled, the trace macro has a significant
+// performance cost so we disable it in release builds.
+macro_rules! trace {
+    ($($tt:tt)*) => {
+        if cfg!(debug_assertions) {
+            ::log::trace!($($tt)*);
+        }
```
bjorn3, line 21: Doesn't log have a feature to specify the minimum log level at compile time and then compile out every log below this level?

Amanieu: It does but:
- it's disabled by default.
- this also disables uses of `trace!` outside of regalloc2.

I personally think that regalloc2's use of trace is just way too verbose to be enabled in release builds.

bjorn3: Trace is meant for things that are extremely verbose. I think trace should almost always be disabled in release builds. You can use the `release_max_level_debug` feature of log for that.

https://docs.rs/log/latest/log/enum.Level.html

> Trace
> 
> The "trace" level.
>
> Designates very low priority, often extremely verbose, information.

Amanieu: If I have a 100k functions to compile then logging ~10 lines per function is already very verbose. This is what I am currently using `trace` for. Having regalloc2 spew out thousands of lines per function is just unreasonable.

bjorn3: I guess so.

src/ion/data_structures.rs
```diff
@@ -544,3 +599,17 @@ pub struct Stats {
     pub halfmoves_count: usize,
     pub edits_count: usize,
 }
+
+// Helper function for generating sorting keys. The order of arguments is from
+// the most significant field to the least significant one.
+//
+// These work best when the fields are stored in reverse order in memory so that
+// they can be loaded with a single u64 load.
```
cfallin, line 607: s/loaded with a single u64 load/loaded with a single u64 load on little-endian machines/

(makes it a bit clearer why this optimization works as well)

src/lib.rs
cfallin: The `trace!()` uses in this crate are more-or-less "stuff I needed to printf-debug" and are definitely verbose; I find it to be the best way to debug very complex algorithmic code, and leave it in so I can get it back when encountering a new fuzzbug or whatever, but it definitely doesn't have to pollute trace-level output globally.

Maybe we could put this under a specific Cargo feature (`detailed-trace` or somesuch) that's off by default? That would keep me happy (rebuild with feature when debugging) while removing this from users' trace output.

cfallin (approved):
This is fantastic work -- thank you so much for finding these optimizations! I find myself both elated that there was the additional performance hiding in in plain sight, and frustrated I didn't see these bits in my earlier optimizations. All the more reason to be thankful others are using and improving regalloc2 as well :-)

A comment below on the `trace!()` situation but otherwise this looks good to merge.

Amanieu:
If we could somehow (safely!) get rid of most bound checks on the vectors then it might even be possible to get a 2x speedup!

cfallin:
> If we could somehow (safely!) get rid of most bound checks on the vectors then it might even be possible to get a 2x speedup!

I'm actually pretty curious about this; have you observed bounds checks in particular showing up in any profiling?

It seems it would be technically possible if we build a wrapper around `Vec` and use it everywhere, and just write a very-unsafe `Index`/`IndexMut` impl. If it's 5% or something then the tradeoff isn't there IMHO, but significant speedups might change the balance, as long as we have an aggressive fuzzing effort to catch any would-be OOB accesses reasonably quickly...

Amanieu:
I did some tests with a special build of rustc that disables slice bounds checks. There's an ~8% reduction in instruction count and ~4% reduction in cycles, so it's probably not worth it. 

### With bounds checks

```
         11,710.57 msec task-clock                #    0.997 CPUs utilized          
               118      context-switches          #   10.076 /sec                   
                20      cpu-migrations            #    1.708 /sec                   
            29,231      page-faults               #    2.496 K/sec                  
    49,672,278,210      cycles                    #    4.242 GHz                    
   108,557,548,161      instructions              #    2.19  insn per cycle         
    20,608,129,422      branches                  #    1.760 G/sec                  
       573,488,342      branch-misses             #    2.78% of all branches        
```

### Without bounds checks

```
         11,231.94 msec task-clock                #    1.000 CPUs utilized          
                37      context-switches          #    3.294 /sec                   
                16      cpu-migrations            #    1.425 /sec                   
            29,121      page-faults               #    2.593 K/sec                  
    48,092,115,263      cycles                    #    4.282 GHz                    
   100,507,332,727      instructions              #    2.09  insn per cycle         
    18,066,199,836      branches                  #    1.608 G/sec                  
       577,440,798      branch-misses             #    3.20% of all branches        
```

cfallin:
@Amanieu thanks for the data! IMHO, a 4% cycle-count ( ~= 4% wallclock time) reduction is probably not worth taking on the risk of unchecked indexing, but it's good to know what it is worth if we ever do need it.

PR #26, opened 2022-01-20
Rework checker to not require DefAlloc by tracking all vregs on each alloc.

cfallin:
The symbolic checker currently works by tracking a single symbolic vreg label for each "alloc" (physical register or stack slot). On definition of a vreg into an alloc, the label is updated to the new vreg's name.

This worked quite well when the regalloc was simpler, but in the presence of redundant move elimination, it started to become apparent that the analysis has a shortcoming: when multiple vregs have the same value, and the regalloc has deduced this, it can make use of an alloc that is labeled with one vreg but use it as another vreg and the checker will not validate the use.

In other words, the regalloc became smart enough to avoid emitting unnecessary moves, but the checker was relying on those moves to know the most up-to-date symbolic name for a value in a physical location. In a sense, a register or stackslot can contain *both* vreg1 and vreg2, and the regalloc can use it as either.

The stopgap measure of emitting more DefAllocs as part of the redundant move elimination never quite sat right with me. It works, but it's asking too much of the regalloc to prove why its moves are correct. We should rely less on the regalloc and on complex built-just-for-the-checker plumbing; we should instead improve the checker so that it can prove on its own that the result is correct.

This PR modifies the checker so that its basic abstraction for an alloc's value is a *set* of virtual register labels, rather than just one. The transfer function is a little more complex, but manageable: a move keeps the old label(s) and adds a new one; redefining a vreg into one alloc needs to remove that vreg label from all other alloc's sets.

This completely removes the need for metadata from the regalloc (!); all we need is the original program (pre-alloc, with vregs), the set of allocations, and the set of inserted moves, and we can validate the result. This should mean that we trust our checker-validated allocation results more, and should result in less complexity and maintenance going forward if we improve the allocator further.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -45,11 +45,13 @@
 //!
 //! The analysis lattice is:
 //!
+//! ```plain
 //!                      Top (V)
 //!                         |
 //!                        𝒫(V)   // the Powerset of the set of virtual regs
 //!                         |
 //!                 Bottom ( ∅ )     // the empty set
+//! ```
 //!
 //! and the lattice ordering relation is the subset relation: S ≤ U
 //! iff S ⊆ U. The lattice meet-function is intersection.
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -743,7 +743,6 @@ impl<'a, F: Function> Checker<'a, F> {
                     .unwrap()
                     .push(CheckerInst::Move { into: to, from });
             }
-            _ => {}
         }
     }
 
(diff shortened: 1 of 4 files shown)


cfallin:
@Amanieu would you be willing to review this?

Fuzzer status: it Hasn't Crashed Yet(tm) after ~1h of CPU-time; I'll let it run overnight to see if that holds.

cfallin:
(CI build failure is rustdoc not liking my fancy Unicode symbols in the doc-comment; will figure out how to suppress that...)

cfallin:
Addressed all comments, thanks!

Amanieu (approved):

PR #27, opened 2022-03-03
Support for debug-labels.

cfallin:
If the client adds labels to vregs across ranges of instructions in the
input program, the regalloc will provide metadata in the `Output` that
describes the `Allocation`s in which each such vreg is stored for those
ranges. This allows the client to emit debug metadata telling a debugger
where to find program values at each point in the program.

This API surface had already been sketched, but no implementation
had existed (the returned debug-label info was always empty) prior
to this PR.

Developed as part of the regalloc2 integration in Cranelift. This is not
strictly needed by Cranelift -- currently it does a post-analysis to get
debug info -- but this will let us eliminate that post-analysis, and simplify
other aspects of the integration as well.

I've added debug-label generation to the fuzzer as well. The checker
doesn't validate that the labels correspond to reality, but (i) that would
be a fairly nontrivial change to the checker, (ii) the actual metadata
emission is fairly straightforward, and (iii) the only tricky property -- that
we're emitting metadata for ranges that are actually overlapping the
allocation we think they are -- is checked by a `debug_assert!`.

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -336,10 +336,6 @@ pub struct Env<'a, F: Function> {
     pub liveouts: Vec<IndexSet>,
     pub blockparam_outs: Vec<BlockparamOut>,
     pub blockparam_ins: Vec<BlockparamIn>,
-    /// Blockparam allocs: block, idx, vreg, alloc. Info to describe
-    /// blockparam locations at block entry, for metadata purposes
-    /// (e.g. for the checker).
-    pub blockparam_allocs: Vec<(Block, u32, VRegIndex, Allocation)>,
 
     pub ranges: Vec<LiveRange>,
     pub bundles: Vec<LiveBundle>,
@@ -390,6 +386,7 @@ pub struct Env<'a, F: Function> {
     pub inst_alloc_offsets: Vec<u32>,
     pub num_spillslots: u32,
     pub safepoint_slots: Vec<(ProgPoint, Allocation)>,
+    pub debug_locations: Vec<(u32, ProgPoint, ProgPoint, Allocation)>,
 
     pub allocated_bundle_count: usize,
 
@@ -595,7 +592,6 @@ pub struct Stats {
     pub spill_bundle_reg_success: usize,
     pub blockparam_ins_count: usize,
     pub blockparam_outs_count: usize,
-    pub blockparam_allocs_count: usize,
     pub halfmoves_count: usize,
     pub edits_count: usize,
 }
--- a/src/ion/moves.rs
+++ b/src/ion/moves.rs
@@ -179,6 +179,25 @@ impl<'a, F: Function> Env<'a, F> {
         }
 
         let debug_labels = self.func.debug_value_labels();
+        #[cfg(debug_assertions)]
+        {
+            // Check the precondition that the debug-labels list is
+            // sorted, and that ranges are non-overlapping.
+            let mut last = None;
+            for tuple in debug_labels {
+                if let Some(last) = last {
+                    // Precondition 1: sorted.
+                    debug_assert!(last <= tuple);
+                    // Precondition 2: ranges for a given vreg are
+                    // non-overlapping. Note that `from`s are
+                    // inclusive and `to`s are exclusive.
+                    let &(last_vreg, _, last_to, _) = last;
+                    let &(tuple_vreg, tuple_from, _, _) = tuple;
+                    debug_assert!(last_vreg < tuple_vreg || last_to <= tuple_from);
+                }
+                last = Some(tuple);
+            }
+        }
 
         let mut half_moves: Vec<HalfMove> = Vec::with_capacity(6 * self.func.num_insts());
         let mut reuse_input_insts = Vec::with_capacity(self.func.num_insts() / 2);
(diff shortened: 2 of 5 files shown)


cfallin:
cc @Amanieu if you'd like to review as well.

src/ion/data_structures.rs
```diff
@@ -336,10 +336,6 @@ pub struct Env<'a, F: Function> {
     pub liveouts: Vec<IndexSet>,
     pub blockparam_outs: Vec<BlockparamOut>,
     pub blockparam_ins: Vec<BlockparamIn>,
-    /// Blockparam allocs: block, idx, vreg, alloc. Info to describe
-    /// blockparam locations at block entry, for metadata purposes
-    /// (e.g. for the checker).
-    pub blockparam_allocs: Vec<(Block, u32, VRegIndex, Allocation)>,
```
cfallin, line 342: I also did this driveby removal and a few under `cfg(feature = "checker")` below as I noticed I hadn't fully removed the machinery that the old checker design required.

src/ion/moves.rs
```diff
@@ -519,6 +502,73 @@ impl<'a, F: Function> Env<'a, F> {
...
+                    //
+                    // We search by the *end* of vale-label range as
+                    // the key in the slice (`tuple.2`), with the
+                    // *start* of this vreg range as the search key,
+                    // so we will find the first value-label range
+                    // with an end equal to or less than this range's
+                    // start. As long as we are careful to exclude the
+                    // just-adjacent case (end == start)
+                    // below, this will give us the first value-label
+                    // that possibly exists in this vreg range.
```
Amanieu, line 522: I personally find it much easier to use `binary_search_by` and a predicate that never returns `Ordering::Equal`. Then you can just `unwrap_err` the result and get the index that is after the previous instruction but before the one that you are looking for.

src/lib.rs
```diff
@@ -958,7 +958,14 @@ pub trait Function {
...
     /// the entire requested ranges.
-    fn debug_value_labels(&self) -> &[(Inst, Inst, VReg, u32)] {
+    ///
+    /// The instruction indices imply a program point just *before*
+    /// the instruction.
+    ///
+    /// Preconditions: we require this slice to be sorted in
+    /// lexicographic order (i.e., first by vreg, then by instruction
+    /// index), and we require the ranges to be non-overlapping.
+    fn debug_value_labels(&self) -> &[(VReg, Inst, Inst, u32)] {
```
Amanieu, line 968: I'm having trouble seeing how passing a range along with a `VReg` is useful here. Why would a single `VReg` have more than one debug label in practice?

A much simpler API would be:
```rust
fn debug_value_label(&self, vreg: VReg) -> Option<u32> { ... }
```

cfallin: Yeah, I agree that would be a lot simpler. Unfortunately at least the debug support in Cranelift allows for labels to be attached to `Value`s only for certain ranges, so we need to carry this through. I believe that this comes up for example when a Wasm-level local is represented by multiple SSA values (due to control flow and multiple assignments); any given label-to-vreg mapping needs bounds that are carried through to the DWARF. And, the `cranelift-frontend` API has `FunctionBuilder::set_val_label` which allows attaching an arbitrary number of labels to a value, starting at arbitrary program points. We could simplify that first, but IMHO I'd rather allow for the more general approach (especially if relatively low-cost) at the regalloc API level; especially when we can't know what other users might require.

src/ion/moves.rs
```diff
@@ -519,6 +502,73 @@ impl<'a, F: Function> Env<'a, F> {
...
+                // Scan debug-labels on this vreg that overlap with
+                // this range, producing a debug-info output record
+                // giving the allocation location for each label.
+                if !debug_labels.is_empty() {
+                    // Do a binary search to find the start of any
+                    // relevant labels. Recall that we require
+                    // debug-label requests to be sorted by vreg then
+                    // by inst-range.
+                    //
+                    // We search by the *end* of vale-label range as
```
Amanieu, line 514:
```suggestion
                    // We search by the *end* of value-label range as
```

src/ion/moves.rs
Amanieu: See the use of `binary_search_by` in `block_insts_and_edits` for an example of what I am describing.

src/ion/moves.rs
```diff
@@ -519,6 +502,73 @@ impl<'a, F: Function> Env<'a, F> {
...
+                        if label_from >= range.to {
+                            break;
+                        }
+                        if label_vreg.vreg() != vreg.index() {
+                            break;
+                        }
+                        if label_from == label_to {
+                            continue;
+                        }
+                        assert!(label_from < label_to);
```
Amanieu, line 549:
```suggestion
                        debug_assert!(label_from < label_to);
```
I replaced all uses of `assert!` with `debug_assert!` in #25. While this particular case is unlikely to have any impact on performance, it's probably a good rule of thumb to always prefer debug-only assertions unless there is a safety invariant involved.

src/ion/moves.rs
cfallin: Done, thanks!

src/ion/moves.rs
cfallin: Switched to that style, thanks!

src/lib.rs
```diff
@@ -958,7 +958,14 @@ pub trait Function {
...
     /// subset of the requested ranges -- if the value is not live for
     /// the entire requested ranges.
-    fn debug_value_labels(&self) -> &[(Inst, Inst, VReg, u32)] {
+    ///
+    /// The instruction indices imply a program point just *before*
+    /// the instruction.
+    ///
+    /// Preconditions: we require this slice to be sorted in
+    /// lexicographic order (i.e., first by vreg, then by instruction
+    /// index), and we require the ranges to be non-overlapping.
```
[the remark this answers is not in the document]
cfallin: Updated!

src/ion/moves.rs
cfallin: Done!

cfallin:
Thanks, updated as per your feedback!

Amanieu (approved):

PR #28, opened 2022-03-04
Simplify pinned-vreg API: don't require slice of all pinned vregs.

cfallin:
Previously, we kept a bool flag `is_pinned` in the `VRegData`, and we
required a `&[VReg]` of all pinned vregs to be provided by
`Function::pinned_vregs()`. This was (I think) done for convenience, but
it turns out not to really be necessary, as we can just query
`is_pinned_vreg` where needed (and in the likely implementation, e.g. in
Cranelift, this will be a `< NUM_PINNED_VREGS` check that can be
inlined). This adds convenience for the embedder (the main benefit), and
also reduces complexity, removes some state, and avoids some work
initializing the regalloc state for a run.

### DIFF
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -503,392 +503,388 @@ impl<'a, F: Function> Env<'a, F> {
                         debug_assert_eq!(dst.kind(), OperandKind::Def);
                         debug_assert_eq!(dst.pos(), OperandPos::Late);
 
-                        // If both src and dest are pinned, emit the
-                        // move right here, right now.
-                        if self.func.is_pinned_vreg(src.vreg()).is_some()
-                            && self.func.is_pinned_vreg(dst.vreg()).is_some()
-                        {
-                            // Update LRs.
-                            if !live.get(src.vreg().vreg()) {
-                                let lr = self.add_liverange_to_vreg(
-                                    VRegIndex::new(src.vreg().vreg()),
-                                    CodeRange {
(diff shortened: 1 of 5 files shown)


src/ion/liveranges.rs
```diff
@@ -509,8 +505,8 @@ impl<'a, F: Function> Env<'a, F> {
 
                         // If both src and dest are pinned, emit the
                         // move right here, right now.
-                        if self.vregs[src.vreg().vreg()].is_pinned
-                            && self.vregs[dst.vreg().vreg()].is_pinned
+                        if self.func.is_pinned_vreg(src.vreg()).is_some()
+                            && self.func.is_pinned_vreg(dst.vreg()).is_some()
```
Amanieu, line 509: This would look better as a `match`:
```rust
match (self.func.is_pinned_vreg(src.vreg()), self.func.is_pinned_vreg(dst.vreg())) {
    (Some(src_preg), Some(dst_preg)) => { ... }
    (Some(preg), None) | (None, Some(preg))  => { ... }
    (None, None) => { ... }
}
```

Amanieu (commented):
I like the simplification!

src/ion/liveranges.rs
cfallin: Done!

cfallin:
Updated based on the suggestion, thanks!

Amanieu (approved):

PR #30, opened 2022-03-17
Allow pinned-vregs to be implicit liveins.

cfallin:
Previously, the regalloc required all liveins to be defined by a
pseudoinstruction at the start of the function body. The regalloc.rs
compatibility shim did this, but it's slightly inconvenient when using
the API directly. This change allows pinned vregs to be implicit liveins
to the function body instead.

Developed during Cranelift integration (bytecodealliance/wasmtime#3942).


Amanieu (approved):
This should be documented on `is_pinned_vreg` in `lib.rs`.

PR #31, opened 2022-03-17
Take vreg regclasses from the Function interface explicitly.

cfallin:
Currently, we infer this from the first def we see of a vreg, but this
is somewhat brittle, as I have found when debugging the integration of
regalloc2 with Cranelift. It would be better to just obtain the register
class for each vreg directly from the embedder. This is a small change
that should be a direct type lookup in most embedders.

### DIFF
--- a/src/fuzzing/func.rs
+++ b/src/fuzzing/func.rs
@@ -140,6 +140,10 @@ impl Function for Func {
         self.num_vregs
     }
 
+    fn vreg_regclass(&self, _: usize) -> RegClass {
+        RegClass::Int
+    }
+
     fn spillslot_size(&self, regclass: RegClass) -> usize {
         match regclass {
             RegClass::Int => 1,
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -14,7 +14,7 @@
 
 use super::{
     CodeRange, Env, InsertMovePrio, LiveBundle, LiveBundleIndex, LiveRange, LiveRangeFlag,
-    LiveRangeIndex, LiveRangeKey, LiveRangeListEntry, LiveRangeSet, PRegData, PRegIndex, RegClass,
+    LiveRangeIndex, LiveRangeKey, LiveRangeListEntry, LiveRangeSet, PRegData, PRegIndex,
     SpillSetIndex, Use, VRegData, VRegIndex, SLOT_NONE,
 };
 use crate::indexset::IndexSet;
@@ -112,7 +112,7 @@ impl<'a, F: Function> Env<'a, F> {
         // Create VRegs from the vreg count.
         for idx in 0..self.func.num_vregs() {
             // We'll fill in the real details when we see the def.
-            let reg = VReg::new(idx, RegClass::Int);
+            let reg = VReg::new(idx, self.func.vreg_regclass(idx));
             self.add_vreg(
                 reg,
                 VRegData {
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -452,7 +452,8 @@ impl<'a, F: Function> Env<'a, F> {
 
             // Create vreg data for blockparams.
             for param in self.func.block_params(block) {
-                self.vreg_regs[param.vreg()] = *param;
+                // Make sure the VReg's RegClass is correct.
+                debug_assert_eq!(self.vreg_regs[param.vreg()], *param);
                 self.vregs[param.vreg()].blockparam = block;
             }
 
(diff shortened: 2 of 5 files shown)


Amanieu:
Since the class is now available separately, could `VReg` be changed to just contain an index and nothing else? This would allow eliminating the separate `VRegIndex` type and the conversions between `VReg` and `VRegIndex`.

cfallin:
> Since the class is now available separately, could `VReg` be changed to just contain an index and nothing else? This would allow eliminating the separate `VRegIndex` type and the conversions between `VReg` and `VRegIndex`.

I briefly dove into this, but I think it's probably a bit too disruptive -- there are places at least in Cranelift where we have a VReg but not the context that would implicitly give its class, so it's fairly useful to the embedder at least have a type (VReg) that bundles both the index and regclass, I think.

Amanieu:
I think this PR is a half-measure as it is: we should either keep things as they are with the register class inside `VReg` or completely separate the vreg index from its class. I slightly prefer the second approach since it avoids the potential confusion of having two VRegs with the same index but different classes.

cfallin:
So, there are issues with both other approaches (not insurmountable, but problematic):

- Taking vreg types only from vregs, and not from a separate source-of-truth (this new API), is brittle -- I hit some bugs when bringing things up where LRs/spillslots were getting the wrong class, because they were created first with `Int` by default before we saw a def. This could be solved by adding a separate pre-scan pass that collects all the regclasses it can, but this seems like a weird and indirect way of communicating types.

- Getting types separately from a context is conceptually very clean, but then runs into issues where the context is not available. It's useful to be able to have the type of the vreg as part of the `VReg`. We could alternatively thread the context everywhere it now needs to go (possible, but annoying refactor, and perhaps some borrow-checker pain when we have mut borrows on the func) or we could create a new container type that wraps the `VReg` (now just an index) and the type when we need to carry both; but then we've just reinvented the current `VReg`.

  This is also likely a performance-relevant detail: a lookup in a separate type table is going to be slower than taking the LSB off the u32.

Happy to discuss further but this is definitely a well-explored corner of the design space at this point, deeply baked into a lot of the code, and I want to be careful about following conceptual purity at the cost of real pragmatic impacts...

cfallin:
I'll add too that it would definitely make sense to assert that the type-bit in each VReg that we encounter matches what the `Func` tells us -- I'd be happy to add a `debug_assert` for that, to catch any bugs.

fitzgen:
> This is also likely a performance-relevant detail: a lookup in a separate type table is going to be slower than taking the LSB off the u32.

Given this point, I guess I don't understand the reasoning for changing the interface to take the register class at all, and not always using the vreg itself (which is my understanding of the current state of things before this PR). It seems like we are introducing a second source of truth and, given that performance argument, are unlikely to make this second source of truth the sole source of truth. So it seems like this PR is increasing the possibility of bugs (the debug asserts in the last commit are a hint here) and it isn't clear to me what benefits we are getting for taking on that downside.

Amanieu:
Also regarding performance, we currently already perform lookups in a separate table in regalloc2 when converting a `VRegIndex` to a `VReg` via `vreg_regs`. We might be able to actually improve performance by eliminating these conversions, which would make `VRegIndex` and `VReg` exactly equivalent.

cfallin:
> Given this point, I guess I don't understand the reasoning for changing the interface to take the register class at all, and not always using the vreg itself (which is my understanding of the current state of things before this PR). It seems like we are introducing a second source of truth and, given that performance argument, are unlikely to make this second source of truth the sole source of truth. So it seems like this PR is increasing the possibility of bugs (the debug asserts in the last commit are a hint here) and it isn't clear to me what benefits we are getting for taking on that downside.

In principle I would agree, but funnily enough this PR actually arose as a bugfix due to an issue caused by the current (pre-PR) design. The basic issue is that we need to know the regclass of a vreg at some point to set up some other state (spill bundle, etc). The state of `main` right now is that we initially assume `Int`, and then every time we see a vreg while scanning code during liverange analysis, we observe its class and change the class we recorded in our own data structures accordingly.

That's error-prone -- the bug was that I had missed one spot where we observe vregs (moves, IIRC).

Unfortunately we also have the perf considerations, and the deeply embedded assumptions in Cranelift at least that a register carries its class. So making `VRegIndex` and `VReg` equivalent is a nonstarter: it would require another 1K-line refactor (potentially nontrivial, as we would have to split out a "type context" and weave it into places where we otherwise have nothing but the regs in helpers, or have a `&mut self` borrow outstanding), or else would force me to just embed the bit again at the Cranelift level.

Given that (i) the `VReg` needs the class, and (ii) we also need the class for our other data structures, the options are to either get the "update on first sight" logic right in all cases, or to add a second pre-pass that scans all vregs (implying a second trip through cache for the `Operand` array), or to just ask the embedder. The last option (this PR) feels like the least bad option to me, but other ideas are welcome...

fitzgen:
Look forward to seeing that -- thanks for your patience!

cfallin:
Superseded by #35; closing this PR.

PR #34, opened 2022-03-18
Generalize debug-info support a bit.

cfallin:
Previously, debug value-label support required each vreg to have a
disjoint sequence of instruction ranges, each with one label.

Unfortunately, it's entirely possible for multiple values at the program
level to map to one vreg at the IR level, leading to multiple labels.

This PR generalizes the debug-info generation support to allow for
arbitrary (label, range, vreg) tuples, as long as they are sorted by
vreg, with no other requirements. The lookup is a little more costly
when we generate the debuginfo, but in practice we shouldn't have more
than a *few* debug value labels per vreg, so in practice the constants
should be small.

### DIFF
--- a/src/ion/moves.rs
+++ b/src/ion/moves.rs
@@ -16,7 +16,6 @@ use super::{
     Env, InsertMovePrio, InsertedMove, LiveRangeFlag, LiveRangeIndex, RedundantMoveEliminator,
     VRegIndex, SLOT_NONE,
 };
-
 use crate::ion::data_structures::{BlockparamIn, BlockparamOut, CodeRange, PosWithPrio};
 use crate::moves::ParallelMoves;
 use crate::{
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -962,9 +962,7 @@ pub trait Function {
     /// The instruction indices imply a program point just *before*
     /// the instruction.
     ///
-    /// Preconditions: we require this slice to be sorted in
-    /// lexicographic order (i.e., first by vreg, then by instruction
-    /// index), and we require the ranges to be non-overlapping.
+    /// Precondition: we require this slice to be sorted in by vreg.
```
Amanieu, line 965:
```suggestion
    /// Precondition: we require this slice to be sorted by vreg.
```

src/ion/moves.rs
```diff
@@ -522,58 +502,38 @@ impl<'a, F: Function> Env<'a, F> {
...
-                    // relevant labels. Recall that we require
-                    // debug-label requests to be sorted by vreg then
-                    // by inst-range, with non-overlapping ranges, as
-                    // preconditions (which we verified above).
+                    // labels for this vreg. Recall that we require
+                    // debug-label requests to be sorted by vreg as a
+                    // precondition (which we verified above).
                     let start = debug_labels
-                        .binary_search_by(|&(label_vreg, _label_from, label_to, _label)| {
+                        .binary_search_by(|&(label_vreg, _label_from, _label_to, _label)| {
```
Amanieu, line 509: Not strictly related to this PR, but I recently found out about the [`partition_point`](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.partition_point) method which may be more appropriate here.

Amanieu (approved):

PR #35, opened 2022-03-29
Record vreg classes explicitly during liverange pass.

cfallin:
This resolves an issue seen when the source program uses multiple
regclasses (Int and Float): in some cases, the logic that grabs the
vregs and retains them (with class) in `vreg_regs` missed a register and
we had a class mismatch. This occurred because data structures were
initialized assuming `Int` regclass at first.

This PR instead removes the `vreg_regs` array, stores the class
explicitly as an `Option<RegClass>` in the `VRegData`, and provides a
`Env::vreg()` method that reconstitutes a `VReg` given its index and its
observed class. We "observe" the class of every vreg seen during the
liveness pass (and we assert that every occurrence of the vreg index has
the same class). In this way, we still have a single source-of-truth for
the vreg class (the mention of the vreg itself) and we explicitly
represent the "not observed yet" state (and panic on attempting to use
such a vreg) rather than implicitly taking the wrong class.

Supersedes #31.

Verified with WIP Cranelift integration and Sightglass that this change
has no performance (compile-time or runtime) impact.

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -259,6 +259,8 @@ pub struct VRegData {
     pub ranges: LiveRangeList,
     pub blockparam: Block,
     pub is_ref: bool,
+    // We don't initially know the RegClass until we observe a use of the VReg.
+    pub class: Option<RegClass>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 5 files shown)


fitzgen (approved):
Looks good to me!

PR #37, opened 2022-03-30
Make some improvements to clarity of checker implementation.

### LINKED ISSUE #36: Feedback/questions/review on the checker
Here are my notes, questions, and suggestions while reading over the `regalloc2`
checker. They are pretty raw, feel free to ask for clarification if anything
doesn't make sense.

----------------------

> The analysis lattice is:
>
> ```plain
>                      Top (V)
>                         |
>                        𝒫(V)   // the Powerset of the set of virtual regs
>                         |
>                 Bottom ( ∅ )     // the empty set
> ```

Doesn't `P(V)` already include the empty set and also a top element (V itself)?
Is this because we have an explicit representation of the universe value that
isn't the actual set `V` but just `universe: true`?

---------------------------

> ```rust
> /// Abstract state for an allocation.
> ///
> /// Equivalent to a set of virtual register names, with the
> /// universe-set as top and empty set as bottom lattice element. The
> /// meet-function is thus set intersection.
> #[derive(Clone, Debug, PartialEq, Eq)]
> struct CheckerValue {
>     /// This value is the "universe set".
>     universe: bool,
>     /// The VRegs that this value is equal to.
>     vregs: FxHashSet<VReg>,
> }
> ```

If `universe == true` then the hash set isn't really ever used, right? If my
understanding is correct, then I think this would be slightly cleaner rewritten
as an `enum`:

```rust
enum CheckerValue {
    Universe,
    VRegs(FxHashSet<VReg>),
}
```

Now, not only are we not distracted by having a hash set accessible when
`universe == true` that could potentially lead to bugs if we forget to check
`universe` and only look at the hash set, but we *can't* be wasting heap space
on the hash set's elements when `universe == true` because of the type
definition.

--------------------------

The top level comment is very nice and helpful.

-----------------------------

> ```rust
> /// State that steps through program points as we scan over the instruction stream.
> #[derive(Clone, Debug, PartialEq, Eq)]
> struct CheckerState {
>     top: bool,
>     allocations: FxHashMap<Allocation, CheckerValue>,
> }
> ```

Similar to `CheckerValue` above, I think this could be better rewritten as:

```rust
enum CheckerState {
    Top,
    Allocations(FxHashMap<Allocation, CheckerValue>),
}
```

-----------------------------

That's it! Everything else looks good to me!

cfallin:
This PR makes two changes, both suggested by @fitzgen in #36:

1. It updates the top-level description of the analysis to more
   simply and accurately describe the analysis lattice.

2. It modifies both the `CheckerValue` and `CheckerState` types to be
   enums with separate arms for the top/universe value, and adds helpers
   as appropriate to update the values. There should be no functional
   change; this update just makes the meet-functions and updates more
   clear, and makes a bad state ("top" but with values) unrepresentable.

Closes #36.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -43,18 +43,17 @@
 //!
 //! ## Formal Definition
 //!
-//! The analysis lattice is:
+//! The analysis lattice consists of the elements of 𝒫(V), the
+//! powerset (set of all subsets) of V (the set of all virtual
+//! registers). The ⊤ (top) value in the lattice is V itself, and the
+//! ⊥ (bottom) value in the lattice is ∅ (the empty set). The lattice
+//! ordering relation is the subset relation: S ≤ U iff S ⊆ U. These
+//! definitions imply that the lattice meet-function (greatest lower
+//! bound) is set-intersection.
 //!
-//! ```plain
-//!                      Top (V)
-//!                         |
-//!                        𝒫(V)   // the Powerset of the set of virtual regs
-//!                         |
-//!                 Bottom ( ∅ )     // the empty set
-//! ```
-//!
-//! and the lattice ordering relation is the subset relation: S ≤ U
-//! iff S ⊆ U. The lattice meet-function is intersection.
+//! (For efficiency, we represent ⊤ not by actually listing out all
+//! virtual registers, but by representing a special "top" value, but
+//! the semantics are the same.)
 //!
 //! The dataflow analysis state at each program point (each point
 //! before or after an instruction) is:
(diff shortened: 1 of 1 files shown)


src/checker.rs
```diff
@@ -43,18 +43,17 @@
...
-//!                      Top (V)
-//!                         |
-//!                        𝒫(V)   // the Powerset of the set of virtual regs
-//!                         |
-//!                 Bottom ( ∅ )     // the empty set
-//! ```
-//!
-//! and the lattice ordering relation is the subset relation: S ≤ U
-//! iff S ⊆ U. The lattice meet-function is intersection.
+//! (For efficiency, we reprsent ⊤ not by actually listing out all
```
bjorn3, line 54: *represent

PR #39, opened 2022-04-04
Bump version to 0.1.0 for release.

cfallin:


fitzgen (approved):
I feel like this should probably be `0.1.0` so we can differentiate API breaking changes with pure bug fixes, but I'll leave that up to you to decide.

cfallin:
Good point; updated!

PR #40, opened 2022-04-13
Add serde support for exposed types.

cfallin:
This adds derived `Serialize` and `Deserialize` implementations for
exposed types that describe registers, operands, and related program
inputs; entity indices; and regalloc output types. This allows
serialization of any of the embedder's IR data types that may embed or
build upon regalloc2 types.

These implementations (and the dependency on the `serde` crate itself)
are enabled only when the non-default `enable-serde` feature is
specified.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,6 +29,14 @@ jobs:
     - name: Run tests
       run: cargo test --all --verbose
 
+  # Make sure the code typechecks with non-default features enabled.
+  features:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


bnjbvr (approved):
Nice! Would it make sense to add a small CI step that `cargo check` with this new feature enabled?

cfallin:
> Nice! Would it make sense to add a small CI step that `cargo check` with this new feature enabled?

Yes, good idea; I've added another CI job that tests with `--all-features` (other features weren't checked in CI either).

PR #43, opened 2022-04-16
Support program moves, including pinned vregs, in the checker.

cfallin:
The checker was built to validate programs produced by the fuzzing
testcase generator, which was built before regalloc2 supported special
handling of moves. (In a pure-SSA world, move elision is not needed,
because moves are not needed, and blockparams are the only way of
tying together vregs.)

Due to this, the checker works great for our independent regalloc2
fuzzing setup, but when used on regalloc inputs produced by Cranelift,
cannot prove correctness.

This PR extends the checker's analysis to properly handle "program
moves", which are distinct from regalloc-inserted moves in that they
are present in the original program and hence are semantically
relevant. A program move edits all sets of symbolic vregs at all
allocs, and where the source vreg appears, it inserts the dest vreg as
well. (It also removes the dest vreg from all other sets, since the
old value becomes stale, as is done for other defs.)

Given this, and given some additional checking for moves to/from
pinned vregs, the checker can now be used to fully validate
Cranelift-sourced regalloc2 invocations.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -582,6 +582,32 @@ impl Operand {
         )
     }
 
+    /// Create an `Operand` that designates a use of a vreg and places
+    /// no constraints on its location (i.e., it can be allocated into
+    /// either a register or on the stack).
+    #[inline(always)]
+    pub fn any_use(vreg: VReg) -> Self {
+        Operand::new(
+            vreg,
+            OperandConstraint::Any,
+            OperandKind::Use,
+            OperandPos::Early,
+        )
+    }
+
+    /// Create an `Operand` that designates a def of a vreg and places
+    /// no constraints on its location (i.e., it can be allocated into
+    /// either a register or on the stack).
+    #[inline(always)]
+    pub fn any_def(vreg: VReg) -> Self {
+        Operand::new(
+            vreg,
+            OperandConstraint::Any,
+            OperandKind::Def,
+            OperandPos::Late,
+        )
+    }
+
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -582,6 +582,32 @@ impl Operand {
...
+            OperandConstraint::Any,
+            OperandKind::Use,
+            OperandPos::Early,
+        )
+    }
+
+    /// Create an `Operand` that designates a def of a vreg and places
+    /// no constraints on its location (i.e., it can be allocated into
+    /// either a register or on the stack).
+    #[inline(always)]
```
fitzgen, line 601: Do these _really_ need to be `inline(always)`? That will force inlining even in debug builds, which has a negative impact on compile times. Is plain `inline` good enough here?

fitzgen (approved):
r=me with below addressed

src/lib.rs
cfallin: IIRC this was actually a hotspot when running debug builds at one point. The constructors are bitwise ORs/shifts to put together bitfields, and in most cases have constants for all but vreg, so turn into `vreg | 0xconstant`. that's *probably* faster to compile than a callsite, though I haven't measured. In any case, I'm following the pattern for other ctors here on these new ones, and am hesitant to touch load-bearing perf things without much more testing :-)

cfallin: (I'll go ahead and merge now with this but we can definitely talk more about inlining and compile time if you think it's warranted)

PR #49, opened 2022-05-17
Handle conflict-related liverange splits arising from stack constraints without falling back to spill bundle.

### LINKED ISSUE #48: Use spill-bundles directly for stack-constrained uses
In bytecodealliance/wasmtime#3785, we discovered a case where an unnecessary register-to-register move occurs. The pattern is like:

* Def of value, constrained to a caller-saved register;
* Callsite, clobbering all caller-saves;
* Safepoint, requiring above value on stack.

Currently, we start a liverange for the value, and split just before the safepoint (due to conflict in requirements). We then chop off the tail of the tail of the first liverange and put it in the spillbundle instead. The first liverange gets the register from its constraint; the last gets a stackslot; and the spillbundle gets a *callee-save* register, because it's live across a callsite.

In this particular case, it would be better to move straight to the stackslot before the callsite. We could probably most easily achieve this by (i) putting all stack-constrained uses directly on the spillbundle, and (ii) in such cases, skipping second-chance register allocation for the spillbundle.

This would require some careful evaluation to ensure that it doesn't regress cases where there happens to be a stack constraint in one place but we mostly want to benefit from second-chance allocation for the spillbundle.

A more general solution is for the spillbundle allocation to take hint(s) from the allocations of all the other LRs it is tied to, possibly weighted somehow. However, in this case it would be pretty fragile because one could equally take a hint from either the first LR (constrained to a reg that is clobbered by the callsite) or the last LR; we somehow need to know to prefer the last. And whatever logic we come up with for the above, one could  imagine the inverse case (safepoint first, then callsite, then use in register) where we would want the opposite hinting preference. So, this more general solution likely needs a lot more careful thought re: cost model if it were to be viable.

cfallin:
Currently, we unconditionally trim the ends of liveranges around a split
when we do a split, including splits due to conflicts in a
liverange/bundle's requirements (e.g., a liverange with both a register
and a stack use). These trimmed ends, if they exist, go to the spill
bundle, and the spill bundle may receive a register during second-chance
allocation or otherwise will receive a stack slot.

This was previously measured to reduce contention significantly, because
it reduces the sizes of liveranges that participate in the first-chance
competition for allocations. When a split has to occur, we might as well
relegate the "connecting pieces" to a process that comes later, with a
hint to try to get the right register if possible but no hard connection
to either end.

However, in the case of a split arising from a reg-to-stack /
stack-to-reg conflict, as happens when references are used or def'd as
registers and then cross safepoints, this extra step in the connectivity
(normal LR with register use, then spill bundle, then normal LR with
stack use) can lead to extra moves. Additionally, when one of the LRs
has a stack constraint, contention is far less important; so it doesn't
hurt to skip the trimming step. In fact, it's likely much better to put
the "connecting piece" together with the stack side of the conflict.

Ideally we would handle this with the same move-cost logic we use for
conflicts detected during backtracking, but the requirements-related
splitting happens separately and that logic would need to be generalized
further. For now, this is sufficient to eliminate redundant moves as
seen in e.g. bytecodealliance/wasmtime#3785.

Fixes #48.

### DIFF
--- a/src/ion/process.rs
+++ b/src/ion/process.rs
@@ -18,12 +18,9 @@ use super::{
     Requirement, SpillWeight, UseList,
 };
 use crate::{
-    ion::{
-        data_structures::{
-            CodeRange, BUNDLE_MAX_NORMAL_SPILL_WEIGHT, MINIMAL_BUNDLE_SPILL_WEIGHT,
-            MINIMAL_FIXED_BUNDLE_SPILL_WEIGHT,
-        },
-        requirement::RequirementConflictAt,
+    ion::data_structures::{
+        CodeRange, BUNDLE_MAX_NORMAL_SPILL_WEIGHT, MINIMAL_BUNDLE_SPILL_WEIGHT,
+        MINIMAL_FIXED_BUNDLE_SPILL_WEIGHT,
     },
     Allocation, Function, Inst, InstPosition, OperandConstraint, OperandKind, PReg, ProgPoint,
     RegAllocError,
@@ -395,6 +392,9 @@ impl<'a, F: Function> Env<'a, F> {
         bundle: LiveBundleIndex,
         mut split_at: ProgPoint,
         reg_hint: PReg,
+        // Do we trim the parts around the split and put them in the
+        // spill bundle?
+        trim_ends_into_spill_bundle: bool,
     ) {
         self.stats.splits += 1;
         trace!(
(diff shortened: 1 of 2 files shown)


cfallin:
cc @Amanieu or @uweigand if either if you would like to review!

cfallin:
No observed compile-time or execution-time impacts with Sightglass. (Though there was a slight compile-time regression before the `#[inline(always)]` pragmas were added in the second commit, a few percent on SpiderMonkey.wasm.) Addresses a manually-observed codegen regression in bytecodealliance/regalloc2#3785.

Amanieu (approved):

PR #51, opened 2022-05-21
Remove an explicitly-set-aside scratch register per class.

### LINKED ISSUE #45: Remove the need for dedicated, non-allocatable scratch registers per class
Currently, regalloc2 requires the embedder to set aside one register from each class as a dedicated scratch register, and exclude this from the list of allocatable registers.

The scratch register for a given class allows "parallel moves" to work in all cases: when we have a cycle of moves that semantically happen concurrently on an edge (due to arguments to blockparams, aka φ-nodes), we need one more storage space than the cycle length in order to lower into a sequence of individual moves.

However, we could take one of several other strategies:
- We could look at the `PReg` allocation maps at the given location, finding a `PReg` that is not in use and using it as the scratch for this one cyclic move;
- Or, we could *create* a scratch register by temporarily saving a register that doesn't participate in the cycle into a stackslot;
- Or, we could use a stackslot as the temporary (this is the only possibility if the cycle involves all registers in the class).

Some care must be taken to handle stack-to-stack moves (we currently guarantee that we don't generate these by using the scratch register as well, saving it off temporarily in a sort of two-level scheme).

The benefit of this would be the addition of one register per class, reducing register pressure and improving runtime performance.

cfallin:
Currently, regalloc2 sets aside one register per class, unconditionally,
to make move resolution possible. To solve the "parallel moves problem",
we sometimes need to conjure a cyclic permutation of data among
registers or stack slots (this can result, for example, from blockparam
flow that swaps two values on a loop backedge). This set-aside scratch
register is used when a cycle exists.

regalloc2 also uses the scratch register when needed to break down a
stack-to-stack move (which could happen due to blockparam moves on edges
when source and destination are both spilled) into a stack-to-reg move
followed by reg-to-stack, because most machines have loads and stores
but not memory-to-memory moves.

A set-aside register is certainly the simplest solution, but it is not
optimal: it means that we have one fewer register available for use by
the program, and this can be costly especially on machines with fewer
registers (e.g., 16 GPRs/XMMs on x86-64) and especially when some
registers may be set aside by our embedder for other purposes too. Every
register we can reclaim is some nontrivial performance in large function
bodies!

This PR removes this restriction and allows regalloc2 to use all
available physical registers. It then solves the two problems above,
cyclic moves and stack-to-stack moves, with a two-stage approach:

- First, it finds a location to use to resolve cycles, if any exist. If
  a register is unallocated at the location of the move, we can use it.
  Often we get lucky and this is the case. Otherwise, we allocate a
  stackslot to use as the temp. This is perfectly fine at this stage,
  even if it means that we have more stack-to-stack moves.

- Then, it resolves stack-to-stack moves into stack-to-reg /
  reg-to-stack. There are two subcases here. If there is *another*
  available free physical register, we opportunistically use it for this
  decomposition. If not, we fall back to our last-ditch option: we pick
  a victim register of the appropriate class, we allocate another
  temporary stackslot, we spill the victim to that slot just for this
  move, we do the move in the above way (stack-to-reg / reg-to-stack)
  with the victim, then we reload the victim. So one move (original
  stack-to-stack) becomes four moves, but no state is clobbered.

This PR extends the `moves` fuzz-target to exercise this functionality
as well, randomly choosing for some spare registers to exist or not, and
randomly generating {stack,reg}-to-{stack,reg} moves in the initial
parallel-move input set. The target does a simple symbolic simulation of
the sequential move sequence and ensures that the final state is
equivalent to the parallel-move semantics.

I fuzzed both the `moves` target, focusing on the new logic; as well as
the `ion_checker` target, checking the whole register allocator, and
both seem clean (~150M cases on the former, ~1M cases on the latter).

Fixes #45.

### DIFF
--- a/fuzz/fuzz_targets/moves.rs
+++ b/fuzz/fuzz_targets/moves.rs
@@ -7,30 +7,56 @@
 use libfuzzer_sys::arbitrary::{Arbitrary, Result, Unstructured};
 use libfuzzer_sys::fuzz_target;
 
-use regalloc2::fuzzing::moves::ParallelMoves;
-use regalloc2::{Allocation, PReg, RegClass};
-use std::collections::HashSet;
+use regalloc2::fuzzing::moves::{MoveAndScratchResolver, ParallelMoves};
+use regalloc2::{Allocation, PReg, RegClass, SpillSlot};
+use std::collections::{HashMap, HashSet};
 
 #[derive(Clone, Debug)]
 struct TestCase {
     moves: Vec<(Allocation, Allocation)>,
+    available_pregs: Vec<Allocation>,
 }
 
 impl Arbitrary for TestCase {
     fn arbitrary(u: &mut Unstructured) -> Result<Self> {
-        let mut ret = TestCase { moves: vec![] };
+        let mut ret = TestCase {
+            moves: vec![],
+            available_pregs: vec![],
+        };
         let mut written = HashSet::new();
+        // An arbitrary sequence of moves between registers 0 to 29
+        // inclusive.
         while bool::arbitrary(u)? {
-            let reg1 = u.int_in_range(0..=30)?;
-            let reg2 = u.int_in_range(0..=30)?;
-            if written.contains(&reg2) {
+            let src = if bool::arbitrary(u)? {
+                let reg = u.int_in_range(0..=29)?;
+                Allocation::reg(PReg::new(reg, RegClass::Int))
+            } else {
+                let slot = u.int_in_range(0..=31)?;
+                Allocation::stack(SpillSlot::new(slot, RegClass::Int))
+            };
+            let dst = if bool::arbitrary(u)? {
+                let reg = u.int_in_range(0..=29)?;
+                Allocation::reg(PReg::new(reg, RegClass::Int))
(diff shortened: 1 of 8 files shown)


cfallin:
Benchmark results with Cranelift: `blake3-scalar` gets 5-8% faster, runtimes otherwise unaffected.

```
execution :: cycles :: benchmarks-next/blake3-scalar/benchmark.wasm

  Δ = 26130.32 ± 5035.33 (confidence = 99%)

  new.so is 1.05x to 1.08x faster than old.so!
  old.so is 0.93x to 0.95x faster than new.so!

  [381216 389967.40 419786] new.so
  [407474 416097.72 445626] old.so

execution :: nanoseconds :: benchmarks-next/blake3-scalar/benchmark.wasm

  Δ = 6874.52 ± 1325.00 (confidence = 99%)

  new.so is 1.05x to 1.08x faster than old.so!
  old.so is 0.93x to 0.95x faster than new.so!

  [100317 102615.16 110455] new.so
  [107217 109489.68 117256] old.so
```

fitzgen (approved):
Thanks for the extensive comments, very helpful when reviewing!

PR #54, opened 2022-05-30
Handle conflicting Before and After fixed-reg constraints with a copy.

### LINKED ISSUE #53: Impossible constraints when a def and use have same fixed-reg constraint and use is live-out
regalloc2 crashes with: `'Could not allocate minimal bundle, but the allocation problem should be possible to solve', regalloc2-0.2.0/src/ion/process.rs:1006:17`

It seems to be related to 128-bit multiplication when targeting x86_64. The following is a manually reduced test-case:

```
function %regalloc_crash(i64, i64, i64) {
block0(v0: i64, v1: i64, v2: i64):
    v3 = uextend.i128 v1
    v4 = uextend.i128 v2

    v5 = imul.i128 v3, v4
    store.i128 v5, v0

    store.i64 v1, v0
    return
 }
```

Potentially related to #19 but continues to crash with the latest version.

cfallin:
This fixes #53. Previously, if two operands on an instruction
specified *different* vregs constrained to the same physical register
at the Before (Early) and After (Late) points of the instruction, and
the Before was live downward as well, we would panic: we can't insert
a move into the middle of an instruction, so putting the first vreg in
the preg at Early implies we have an unsolveable conflict at Late.

We can solve this issue by adding some new logic to insert a copy, and
rewrite the constraint. This reuses the multi-fixed-reg-constraint
fixup logic. While that logic handles the case where the *same* vreg
has multiple *different* fixed-reg constraints, this new logic
handles *different* vregs with the *same* fixed-reg constraints, but
at different *program points*; so the two are complementary.

This addresses the specific test case in #53, and also fuzzes cleanly
with the change to the fuzz testcase generator to generate these
cases (which also immediately found the bug).

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -590,18 +590,24 @@ impl CheckerState {
     fn update<'a, F: Function>(&mut self, checkinst: &CheckerInst, checker: &Checker<'a, F>) {
         self.become_defined();
 
-        let default_val = Default::default();
         match checkinst {
             &CheckerInst::Move { into, from } => {
-                let val = self.get_value(&from).unwrap_or(&default_val).clone();
-                trace!(
-                    "checker: checkinst {:?} updating: move {:?} -> {:?} val {:?}",
-                    checkinst,
-                    from,
-                    into,
-                    val
-                );
-                self.set_value(into, val);
+                // Value may not be present if this move is part of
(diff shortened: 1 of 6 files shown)


fitzgen (approved):
LGTM!

PR #59, opened 2022-06-26
Limit split count per original bundle with fallback 1-to-N split.

cfallin:
Right now, splitting a bundle produces two halves. Furthermore, it has
cost linear in the length of the bundle, because the resulting
half-bundles have their requirements recomputed with a new scan, and
because we copy half the use-list over to the tail end sub-bundle.

This works fine when a bundle has a handful of splits overall, but not
when an input has a systematic pattern of conflicts that will require
O(|bundle|) splits (e.g., every Use is constrained to a different fixed
register than the last one). In such a case, we get quadratic behavior.

This PR adds a per-spillset (so, per-original-bundle) counter for
splits, and when it reaches a preset threshold (10 for now), we instead
split directly into minimal bundles along the whole length of the
bundle, putting the regions without uses in the spill bundle.

This basically approximates what a non-splitting allocator would do: it
"spills" the whole bundle to possibly a stackslot, or a second-chance
register allocation at best, via the spill bundle; and then does minimal
reservations of registers just at uses/defs and moves the "spilled"
value into/out of them immediately.

Together with another small optimization, this PR results in a 4x
compilation speedup and 24x memory use reduction on one particularly bad
case with alternating conflicting requirements on a vreg (see
bytecodealliance/wasmtime#4291 for details).


cfallin:
Apologies to whomever reviews this for the hairy innards of the LR/bundle representation and splitting!

It's been fuzzing overnight on my 12-core system without finding anything (after a few initial fuzzbugs yesterday that guided some corner-cases here) so it seems relatively stable...

src/ion/process.rs
```diff
@@ -728,6 +739,221 @@ impl<'a, F: Function> Env<'a, F> {
...
+
+                last_lr = Some(lr);
+                last_bundle = Some(new_bundle);
+                last_inst = Some(u.pos.inst());
+                last_vreg = Some(vreg);
+            }
+
+            // Clear the use-list from the original LR.
+            self.ranges[entry.index.index()].uses.clear();
+            self.ranges[entry.index.index()].uses.shrink_to_fit();
```
alexcrichton, line 900: Stylistically this can probably be also done with `... = Default::default()` since the new collection will have 0 capacity to start out with.

src/ion/process.rs
```diff
@@ -728,6 +739,221 @@ impl<'a, F: Function> Env<'a, F> {
...
+            bundle,
+            reg_hint
+        );
+
+        let mut last_lr: Option<LiveRangeIndex> = None;
+        let mut last_bundle: Option<LiveBundleIndex> = None;
+        let mut last_inst: Option<Inst> = None;
+        let mut last_vreg: Option<VRegIndex> = None;
+
+        for entry_idx in 0..self.bundles[bundle.index()].ranges.len() {
```
alexcrichton, line 766: Random stylistic comment but you can `impl<T> Index<MyLocalType> for [T]` which could help change all the indexing here to `self.bundles[bundle]` (or even `self[bundle]`) rather than `self.bundles[bundle.index()]`

alexcrichton (commented):
Those are some awesome wins! Unfortunately though I'm pretty unfamiliar with the implementation here so it would take me quite some time to get up to speed and able to review this. I had a random few stylistic comments and while at a high level this seems reasonable I'll leave it to @fitzgen to review the internals.

fitzgen (approved):

src/ion/process.rs
```diff
@@ -410,6 +410,15 @@ impl<'a, F: Function> Env<'a, F> {
...
         let spillset = self.bundles[bundle.index()].spillset;
 
+        // Have we reached the maximum split count? If so, fall back
+        // to a "minimal bundles and spill bundle" setup for this
+        // bundle.
+        if self.spillsets[spillset.index()].splits >= MAX_SPLITS_PER_SPILLSET {
+            self.split_into_minimal_bundles(bundle, reg_hint);
+            return;
+        }
+        self.spillsets[spillset.index()].splits += 1;
```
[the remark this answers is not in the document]
cfallin: Done!

src/ion/process.rs
cfallin: Done!

src/ion/process.rs
cfallin: I briefly played with this but it seems we would need wrapper types around the `Vec<T>` to make this work properly. I'd love to have this eventually but it's a bigger refactor (all use-sites across the allocator) than I want to bite off right now; I'll file an issue!

cfallin: #62

PR #64, opened 2022-07-20
Streamline log enablement

bnjbvr:
Two small changes related to logging, that try to simplify reasoning about how logs are enabled a bit:

- `annotate` already has a guard in the beginning of its body, checking if `annotations_enabled` is true. Most uses of this function aren't additionally guarded with `if log::log_enabled!(log::Level::Trace)`, so for consistency and since there's already the `enabled_annotations` guard, I've removed the extra `log_enabled` checks (in two places).
- `dump_state` uses the crate-defined `trace` macro that checks against the `trace-log` feature, but the check on top of `dump_state` uses the `log_enabled` check. Instead I've introduced another macro similar to `log::log_enabled` that follows the same logic as the crate-defined `trace` macro to decide whether to enter the block or not.

### DIFF
--- a/src/ion/moves.rs
+++ b/src/ion/moves.rs
@@ -435,19 +435,17 @@ impl<'a, F: Function> Env<'a, F> {
                                     alloc,
                                 );
                                 #[cfg(debug_assertions)]
-                                {
-                                    if log::log_enabled!(log::Level::Trace) {
-                                        self.annotate(
-                                            self.cfginfo.block_entry[block.index()],
-                                            format!(
-                                                "blockparam-in: block{} to block{}:into v{} in {}",
-                                                from_block.index(),
-                                                to_block.index(),
-                                                to_vreg.index(),
-                                                alloc
-                                            ),
-                                        );
-                                    }
+                                if self.annotations_enabled {
+                                    self.annotate(
+                                        self.cfginfo.block_entry[block.index()],
+                                        format!(
+                                            "blockparam-in: block{} to block{}:into v{} in {}",
+                                            from_block.index(),
+                                            to_block.index(),
+                                            to_vreg.index(),
+                                            alloc
+                                        ),
+                                    );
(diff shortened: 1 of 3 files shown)


src/ion/moves.rs
```diff
@@ -436,18 +436,16 @@ impl<'a, F: Function> Env<'a, F> {
...
-                                                from_block.index(),
-                                                to_block.index(),
-                                                to_vreg.index(),
-                                                alloc
-                                            ),
-                                        );
-                                    }
+                                    self.annotate(
+                                        self.cfginfo.block_entry[block.index()],
+                                        format!(
```
cfallin, line 441: I think this change will mean that in `debug_assertions` builds with logging disabled, we now invoke this `format!()` call and allocate a string where we didn't before? Probably it should be guarded with the `trace_enabled!()` conditional as above...

cfallin (commented):
Thanks! Happy to see these improvements. Just one thing below...

src/ion/moves.rs
bnjbvr: Ah you're right that it will allocate; to keep on using only the same single guard, i think it should be guarded with `annotations_enabled` too. Thanks!

src/ion/moves.rs
```diff
@@ -832,15 +830,10 @@ impl<'a, F: Function> Env<'a, F> {
...
-                            if log::log_enabled!(log::Level::Trace) {
-                                self.annotate(
-                                    ProgPoint::before(inst),
-                                    format!(
-                                        " reuse-input-copy: {} -> {}",
-                                        input_alloc, output_alloc
-                                    ),
-                                );
-                            }
+                            self.annotate(
```
cfallin, line 833: Here as well I think we need a conditional on `annotations_enabled`?

cfallin (approved):
Thanks!

PR #66, opened 2022-07-28
Only record vreg definitions when fuzzing

jameysharp:
The `vreg_def_blockparam` and `vreg_def_inst` fields on CFGInfo are only
used in `validate_ssa`, which in turn is only used in the ssagen fuzz
target.

Since these fields are never read in normal usage, initializing them is
entirely wasted effort. According to valgrind/DHAT, when running
`wasmtime compile` on the Sightglass Spidermonkey benchmark, removing
these fields saves about 100M instructions, 23k heap allocations
totalling 40MiB, and 47MiB of writes to the heap.

I tested the fuzz target for a few minutes, and also tested a release build of wasmtime. So I've built with the fuzzing feature both on and off, and there weren't any build warnings or errors in either case.

### DIFF
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -5,7 +5,7 @@
 
 //! Lightweight CFG analyses.
 
-use crate::{domtree, postorder, Block, Function, Inst, OperandKind, ProgPoint, RegAllocError};
+use crate::{domtree, postorder, Block, Function, Inst, ProgPoint, RegAllocError};
 use smallvec::{smallvec, SmallVec};
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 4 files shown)


cfallin (commented):
Thanks for this!

Rather than fine-grained `cfg`s on individual statements, a pattern that I like to follow for easier maintainability and less ifdef-soup (and one that I picked up from Alex during memfd work on Wasmtime) is to make them more coarse-grained: either alternate implementations of structs that are empty, with empty/dummy methods, or alternate modules.

In the extreme, `SSAInfo` would just move into the `ssa` module entirely; doing a separate pass to fill it in isn't a huge deal, since this only affects the `ssagen`-target fuzzing flow and the SSA check is relatively heavyweight anyway. The only reason it's still intertwined with the `CFGInfo` analysis after this PR's changes appears to be to share the loop over blocks and instructions within a block; those are simple enough that we don't need to deduplicate complexity.

Alternately we could have a separate `SSAInfo` struct that's empty, for `#[cfg(not(feature = "fuzzing"))]`, and an `impl` on it, also cfg'd, that has empty methods. But I prefer the above option actually; there's no reason for `SSAInfo` to live in `cfg.rs` at all.

cfallin (approved):
Ah, this is much cleaner -- thanks a bunch!

PR #67, opened 2022-07-29
Faster SSA validation

jameysharp:
Welp, as you expected in #66 @cfallin, this probably wasn't worth doing. I dunno, maybe it uses limited fuzzing CPU budgets more efficiently? Anyway it was nice to review SSA invariants and get a little more insight into regalloc2 details, so I don't regret the experiment.

### DIFF
--- a/src/fuzzing/ssa.rs
+++ b/src/fuzzing/ssa.rs
@@ -52,6 +52,8 @@ pub fn validate_ssa<F: Function>(f: &F, cfginfo: &CFGInfo) -> Result<(), RegAllo
             }
             *def_block = block;
         }
+
+        let mut saw_branch = false;
         for iix in f.block_insns(block).iter() {
             let operands = f.inst_operands(iix);
             for operand in operands {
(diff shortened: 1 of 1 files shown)


cfallin:
Hmm, unfortunately I don't think this is an improvement; and I think it's worth spelling out some details of that to help guide things in general:

- The original code had two passes, one to find defs' locations and one to check that each use uses only defs from a dominating block. This code has one pass, doing both at once. However, the cleverness required to do that is quite substantial: one has to reason about visit orders and it's not immediately obvious why it should work if the input may not be in SSA. (Once it *is* in SSA, it's a lot easier to reason about def-use direction  and traversal order with the domtree, but the interplay of "we're validating the property" and "we're relying on the property" here is very subtle.) 

- It's not actually clear to me if the single-pass version is faster than the current two-pass approach, given that the version shown here does somewhat complex things in allocating a traversal order array, inverting the domtree, etc. That's a lot of work to go through in order to traverse memory twice rather than once.

- The purpose of this validator is not to be in the hotpath of compilation, and not even to be in the hotpath of the main fuzzer (the symbolic checker of the allocator's results), but only to support an auxiliary fuzz target that validates the function-body generator. For this reason, clarity and correctness of the code is *far* more important than performance: it's part of the "chain of confidence" used to build up confidence in the final translation validation.

While I do very much appreciate any effort at optimization (thank you for playing with this!), I think here the engineering tradeoffs are pretty clear in the other direction. I think for further optimization efforts it would be good to look at profiling results and be guided by where time is spent, rather than theoretical algorithmic improvements. Thanks!

cfallin:
(I'm going to go ahead and close this based on the above reasoning, but happy to discuss further if you'd like!)

jameysharp:
No, that's all entirely reasonable!

PR #68, opened 2022-08-03
validate_ssa: Check def uniqueness first

jameysharp:
This is an alternative to #67, based on things I learned while writing that.

I had trouble understanding `validate_ssa`, so this changes the division
of work to establish one invariant at a time:

- First, check that no VReg is defined in more than one place, and
  record which block it's defined in.
- Second, check that each use refers to a VReg that was defined either
  earlier in the same block, or in a block dominating it.

I'm also enhancing the validation to check that every VReg use is in an
instruction that is strictly later than its def. Previously, an earlier
operand in the same instruction would also be considered a valid def,
but that isn't actually legal in SSA.

### DIFF
--- a/src/fuzzing/ssa.rs
+++ b/src/fuzzing/ssa.rs
@@ -5,30 +5,31 @@
 
 //! SSA-related utilities.
 
-use crate::cfg::CFGInfo;
+use std::collections::HashSet;
 
-use crate::{Block, Function, Inst, OperandKind, RegAllocError};
+use crate::cfg::CFGInfo;
+use crate::{Block, Function, Inst, OperandKind, RegAllocError, VReg};
 
 pub fn validate_ssa<F: Function>(f: &F, cfginfo: &CFGInfo) -> Result<(), RegAllocError> {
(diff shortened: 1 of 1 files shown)


src/fuzzing/ssa.rs
```diff
@@ -37,40 +38,30 @@ pub fn validate_ssa<F: Function>(f: &F, cfginfo: &CFGInfo) -> Result<(), RegAllo
...
                             return Err(RegAllocError::SSA(operand.vreg(), iix));
                         }
                     }
-                    OperandKind::Def => {
-                        if defined[operand.vreg().vreg()] {
-                            return Err(RegAllocError::SSA(operand.vreg(), iix));
-                        }
-                        defined[operand.vreg().vreg()] = true;
-                    }
+                    OperandKind::Def => {}
```
cfallin, line 64: Perhaps a comment here (inside the block) saying that defs are processed below, after all uses are checked?

cfallin (approved):
LGTM, thanks! Just one request for a clarifying comment below.

cfallin (approved):

PR #69, opened 2022-08-03
Remove old tests and reexport libfuzzer from fuzzing module

jameysharp:
* Remove unused regalloc2-test crate
    
    This code doesn't build, and Chris says it's "a really old harness that
    existed prior to building the fuzzing and was used mainly to profile and
    get stats before integration with Cranelift".

* Re-export libfuzzer/arbitrary from fuzzing module
    
    This avoids needing to keep dependencies on `arbitrary` in sync across
    the three different Cargo.toml files in this project.
    
    However, before version 0.4.2, libfuzzer-sys only supported using its
    macros if it was available at the top-level `libfuzzer_sys` path, which
    breaks when re-exporting it. So I'm upgrading to that version (or the
    newest patch release of it).
    
    Upgrading libfuzzer-sys in turn brings in the 1.0 release of the
    arbitrary crate, with a minor API change along the way.


cfallin (approved):
Thanks!

I didn't realize `libfuzzer_sys` re-exported `arbitrary`; that makes things a lot easier here, for sure.

jameysharp:
Yeah, the fuzz targets were already using libfuzzer's re-export of arbitrary, so I just took that a step further.

PR #74, opened 2022-09-09
Some fixes to allow for call instructions to name args, returns, and clobbers with constraints.

### LINKED ISSUE #53: Impossible constraints when a def and use have same fixed-reg constraint and use is live-out

cfallin:
- Allow early-pos uses with fixed regs that conflict with clobbers (which happen at late-pos), in addition to the existing logic for conflicts with late-pos defs with fixed regs.

  This is a pretty subtle issue that was uncovered in #53 for the def case, and the fix here is the mirror of that fix for clobbers. The root cause for all this complexity is that we can't split in the middle of an instruction (because there's no way to insert a move there!) so if a use is live-downward, we can't let it live in preg A at early-pos and preg B != A at late-pos; instead we need to rewrite the constraints and use a fixup move.

  The earlier change to fix #53 was actually a bit too conservative in that it always applied when such conflicts existed, even if the downward arg was not live. This PR fixes that (it's fine for the early-use and late-def to be fixed to the same reg if the use's liverange ends after early-pos) and adapts the same flexibility to the clobbers case as well.

- Reworks the fixups for the def case mentioned above to not shift the def to the Early point. Doing so causes issues when the def is to a reffy vreg: it can then be falsely included in a stackmap if the instruction containing this operand is a safepoint.

- Fixes the last-resort split-bundle-into-minimal-pieces logic from #59 to properly limit a minimal bundle piece to end after the early-pos, rather than cover the entire instruction. This was causing artificial overlaps between args that end after early-pos and defs that start at late-pos when one of the vregs hit the fallback split behavior.

### DIFF
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -987,12 +987,14 @@ impl<'a, F: Function> Env<'a, F> {
                                 // conflicting constraints for the
                                 // same vreg in a separate pass (see
                                 // `fixup_multi_fixed_vregs` below).
-                                if late_def_fixed.contains(&preg) {
+                                if late_def_fixed.contains(&preg)
+                                    || self.func.inst_clobbers(inst).contains(preg)
+                                {
                                     log::trace!(
                                         concat!(
                                             "-> operand {:?} is fixed to preg {:?}, ",
                                             "is downward live, and there is also a ",
-                                            "def at this preg"
+                                            "def or clobber at this preg"
                                         ),
                                         operand,
                                         preg
@@ -1010,43 +1012,11 @@ impl<'a, F: Function> Env<'a, F> {
                                     // We need to insert a reservation
                                     // at the before-point to reserve
                                     // the reg for the use too.
-                                    let range = CodeRange {
-                                        from: ProgPoint::before(inst),
-                                        to: ProgPoint::after(inst),
-                                    };
+                                    let range = CodeRange::singleton(pos);
                                     self.add_liverange_to_preg(range, preg);
 
                                     // Remove the fixed-preg
                                     // constraint from the Use.
-                                    operand_rewrites.insert(
-                                        i,
-                                        Operand::new(
-                                            operand.vreg(),
-                                            OperandConstraint::Any,
-                                            operand.kind(),
-                                            operand.pos(),
-                                        ),
-                                    );
-                                } else if self.func.inst_clobbers(inst).contains(preg) {
-                                    let pos = ProgPoint::before(inst);
-
-                                    // We need to insert a clobber at
-                                    // the before-point to reserve the
-                                    // reg for the use too.
-                                    let range = CodeRange {
-                                        from: ProgPoint::before(inst),
-                                        to: ProgPoint::after(inst),
-                                    };
-                                    self.add_liverange_to_preg(range, preg);
-
-                                    self.multi_fixed_reg_fixups.push(MultiFixedRegFixup {
-                                        pos,
-                                        from_slot: i as u8,
-                                        to_slot: i as u8,
-                                        to_preg: PRegIndex::new(preg.index()),
-                                        vreg: VRegIndex::new(operand.vreg().vreg()),
-                                        level: FixedRegFixupLevel::Initial,
-                                    });
                                     operand_rewrites.insert(
                                         i,
                                         Operand::new(
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -988,6 +988,15 @@ impl<'a, F: Function> Env<'a, F> {
                                 // same vreg in a separate pass (see
                                 // `fixup_multi_fixed_vregs` below).
                                 if late_def_fixed.contains(&preg) {
+                                    log::trace!(
+                                        concat!(
+                                            "-> operand {:?} is fixed to preg {:?}, ",
+                                            "is downward live, and there is also a ",
+                                            "def at this preg"
+                                        ),
+                                        operand,
+                                        preg
+                                    );
                                     let pos = ProgPoint::before(inst);
                                     self.multi_fixed_reg_fixups.push(MultiFixedRegFixup {
                                         pos,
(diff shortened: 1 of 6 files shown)


cfallin:
This one is pretty subtle but perhaps one of @fitzgen, @elliottt, @jameysharp would be willing to review? I'm happy to do a live review / talk through it if that's easier.

cfallin:
Updated to be a little more generic with respect to safepoint handling (not special-casing early defs) as per suggestion by @jameysharp in offline discussion just now. Thanks! Should be ready to review now.

cfallin:
Another update pushed after finding a bug in the above changes while fuzzing locally.

cfallin:
Rebased to current main.

src/ion/liveranges.rs
```diff
@@ -1013,13 +1027,33 @@ impl<'a, F: Function> Env<'a, F> {
...
+                                } else if self.func.inst_clobbers(inst).contains(preg) {
+                                    let pos = ProgPoint::before(inst);
+
+                                    // We need to insert a clobber at
+                                    // the before-point to reserve the
+                                    // reg for the use too.
+                                    let range = CodeRange {
+                                        from: ProgPoint::before(inst),
+                                        to: ProgPoint::after(inst),
+                                    };
```
elliottt, line 1039: Would it be worth adding a `CodeRange::whole_instruction` function for constructing this special case?

src/ion/liveranges.rs
```diff
@@ -1013,13 +1027,33 @@ impl<'a, F: Function> Env<'a, F> {
                                             operand.pos(),
                                         ),
                                     );
+                                } else if self.func.inst_clobbers(inst).contains(preg) {
```
elliottt, line 1030: The branches of these two conditionals seem to be the same except for the trace in the previous one (unless I'm missing something). Would it be worth extracting the common part to a utility function to ensure they stay the same?

cfallin: Excellent observation -- thanks for noticing this! I think that they differed more originally but with some simplifications they actually experienced convergent evolution. Merged into one conditional now!

src/ion/liveranges.rs
cfallin: The range is actually just over the before-point (`from` is inclusive and `to` is exclusive) but I've created `CodeRange::singleton(pos)` for this -- thanks!

src/ion/merge.rs
```diff
@@ -258,23 +271,30 @@ impl<'a, F: Function> Env<'a, F> {
...
+                    if fixed && stack && fixed_def {
                         break;
                     }
                 }
             }
             if fixed {
                 self.bundles[bundle.index()].set_cached_fixed();
             }
+            if fixed_def {
+                self.bundles[bundle.index()].set_cached_fixed_def();
```
elliottt, line 296: Does this flag ever get cleared, or does it follow the parts of the bundle if it's split up?

elliottt: Oh this isn't a flag on the whole bundle, just that range. Never mind :)

cfallin: It's recomputed along with the other flags when the bundle is split, if it ever is (it actually is a flag on the whole bundle!).

cfallin: https://github.com/cfallin/regalloc2/blob/ac8e5f4e22791e7e554c2c8f2ddea28768bbdc78/src/ion/process.rs#L344-L350

src/ion/merge.rs
```diff
@@ -93,6 +95,17 @@ impl<'a, F: Function> Env<'a, F> {
             }
         }
 
+        // Avoid merging if either side has a fixed-reg def: this can
+        // result in an impossible-to-solve allocation problem if
+        // there is a fixed-reg use in the same reg on the same
+        // instruction.
+        if self.bundles[from.index()].cached_fixed_def()
+            || self.bundles[to.index()].cached_fixed_def()
```
elliottt, line 103: If you knew that the fixed def constraints were for the same register, would that change whether or not you needed to bail out here?

cfallin: No, in fact the fuzzbug was exactly this case! The issue is that the way that we do the constraint rewrite adds an anonymous reservation; so this isn't a case of "different fixed def, can't merge" (that is captured by the incompatible-requirement check) but rather "corner case results in impossible problem".

The interesting case is when we have a fixed-use bound to preg P with liverange up to the early point on an inst, a fixed-def bound to preg P (same preg) at the late point on an inst, and that def'd vreg flows into a blockparam arg and over a backedge to the use's vreg. Then we try to merge the two vregs into one bundle (because we try this for every blockparam arg - blockparam def pair). But we rewrote the use to be an "any" use and put an anonymous reservation on preg P at the early-point only. So rejoining the first vreg (the use-side) to the whole bundle and thus requiring it to have the fixed-preg constraint again causes an impossible-to-solve problem.

There may be an even more complex way to rewrite this but I was several levels deep into "introduce fix, get deeper bug" and wanted to bail out of the rabbithole altogether...

elliottt (approved):
This looks good to me! Thanks for answering my questions and for the thorough write-up :)

src/ion/merge.rs
Amanieu: This check is very conservative and is hurting our compiler quite a bit since it makes heavy use of fixed operands. Removing this check gives a 4% code size reduction (I enabled the checker to ensure the result is still valid).

What are the exact requirements that cause the impossible to resolve constraint? Could this check be made more specific?

cfallin: I don't really have details paged in anymore, beyond the paragraph describing a fuzzing-discovered example I wrote above, nor the cycles at the moment to think through it deeply. You're welcome to remove it and fuzz to (re)discover it, and propose a better fix, of course!

PR #77, opened 2022-09-15
Add fixed-non-allocatable operand support

### LINKED ISSUE #3: Come up with a better scheme for fixed-register usages
In many cases, a regalloc2 client will want to generate instructions that refer to fixed registers for uses/defs. There are multiple kinds of these register references: (i) uses/defs of a vreg that needs to be in a certain fixed register just at one instruction (e.g. for a call argument or return value), (ii) uses/defs of a machine register that is not allocatable but is used in an operand slot that can also take allocatable registers. 

The former is well-covered by `OperandConstraint::FixedReg`, but the latter is interesting; for now our answer is to use pinned vregs, but we should have a better way of "passing through" non-allocatable registers, and we should clearly define what the semantics are.

Amanieu:
This allows a non-allocatable `PReg` to be passed on directly to the allocations vector without any liverange tracking from the register allocator. The main intended use case is to support ISA-specific special registers such as a fixed zero register.

Fixes #3

### DIFF
--- a/fuzz/fuzz_targets/ion_checker.rs
+++ b/fuzz/fuzz_targets/ion_checker.rs
@@ -22,6 +22,7 @@ impl Arbitrary<'_> for TestCase {
                 &Options {
                     reused_inputs: true,
                     fixed_regs: true,
+                    fixed_nonallocatable: true,
                     clobbers: true,
                     control_flow: true,
                     reducible: false,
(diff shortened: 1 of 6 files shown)


Amanieu:
I got a 6% speedup in compilation time by replacing my uses of the zero register with `Operand::fixed_nonallocatable`. I was previously defining the zero register using `Operand::reg_fixed_def` on the entry instruction, which was subtly incorrect since regalloc could sometimes emit a move that would write to the zero register.

cfallin (approved):
Looks reasonable to me; thanks!

PR #78, opened 2022-09-15
Handle fixed stack slots in the move resolver

Amanieu:
Fixed stack slots are treated as `PReg`s by most of the register allocator, but need some additional handling the move resolver to avoid generating stack-to-stack moves.


cfallin (approved):
This looks reasonable as well; thanks!

PR #80, opened 2022-09-20
Remove register class from `SpillSlot`

### LINKED ISSUE #79: Type confusion between `Int` and `Float` in the move resolver
This only happens when both `Int` and `Float` classes have the same spillslot size.

The root of the problem is that a single spillslot can be used (separately) to hold values of either class. This is because spill slot allocation only looks at `slots_by_size` and ignores the register class of the spillset. The class of the `SpillSlot` seems to be whichever spillset first allocates this spillslot ([source](https://github.com/bytecodealliance/regalloc2/blob/aeef47a06b88be80a37d160bf21632b7dc0a95a4/src/ion/spill.rs#L168)).

This seems to be intentional since the move resolution code specifically ignores the register class on spillslots ([1](https://github.com/bytecodealliance/regalloc2/blob/aeef47a06b88be80a37d160bf21632b7dc0a95a4/src/ion/moves.rs#L67) [2](https://github.com/bytecodealliance/regalloc2/blob/aeef47a06b88be80a37d160bf21632b7dc0a95a4/src/ion/moves.rs#L1113) [3](https://github.com/bytecodealliance/regalloc2/blob/main/src/ion/moves.rs#L974)).

However this causes problems when constructing the move lists because the register class is obtained from [`m.from_alloc.class()`](https://github.com/bytecodealliance/regalloc2/blob/aeef47a06b88be80a37d160bf21632b7dc0a95a4/src/ion/moves.rs#L980) which may be incorrect if `from_alloc` is a spillslot (i.e. it won't match the class of the other registers in the move list).

I can see 2 ways we might fix this:
1. Require that a spillslot only be used with a single register class. This may end up increasing stack usage.
2. Completely remove the register class from `SpillSlot`. But then the move resolver needs some way to know which scratch register to use for resolving stack-to-stack moves.

Amanieu:
The register allocator was already allowing moves between spillslots and registers of different classes, so this PR formalizes this by making spillslots independent of register class.

This also fixes #79 by properly tracking the register class of an `InsertedMove` with the `to_vreg` field which turns out to never be `None` in practice. Removing the `Option` allows the register class of the `VReg` to be used when building the per-class move lists.

Fixes #79

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -440,7 +440,7 @@ impl<'a, F: Function> Env<'a, F> {
 #[derive(Clone, Debug)]
 pub struct SpillSlotData {
     pub ranges: LiveRangeSet,
-    pub size: u32,
+    pub slots: u32,
     pub alloc: Allocation,
 }
 
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -440,7 +440,7 @@ impl<'a, F: Function> Env<'a, F> {
 #[derive(Clone, Debug)]
 pub struct SpillSlotData {
     pub ranges: LiveRangeSet,
-    pub class: RegClass,
+    pub size: u32,
     pub alloc: Allocation,
 }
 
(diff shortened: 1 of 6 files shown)


src/ion/data_structures.rs
```diff
@@ -440,7 +440,7 @@ impl<'a, F: Function> Env<'a, F> {
 #[derive(Clone, Debug)]
 pub struct SpillSlotData {
     pub ranges: LiveRangeSet,
-    pub class: RegClass,
+    pub size: u32,
```
cfallin, line 443: can we call this field `slots` to make the units clear (as opposed to e.g. bytes)?

cfallin (approved):
Thanks, this is simple and clean! Just one tiny field-naming request below, otherwise LGTM.

PR #82, opened 2022-09-20
Fix design doc typo

CrockAgile:
Feel free to ignore! I was reading over the design doc and noticed a typo. After the first, I was hoping to find more so the PR wouldn't be so trivial, but alas I didn't find anything else.


cfallin (approved):
Thanks!

PR #85, opened 2022-09-22
Fix fallback-split behavior: trim start of minimal bundle wrt start of original LR.

cfallin:
When a liverange starts at a *late* point of an instruction, and it undergoes the fallback "split into all minimal pieces" transform, we end up creating one minimal bundle that starts at the *early* point of the instruction at the start of the original LR. This can create impossible-to-allocate situations where a fixed-constraint LR overlaps another constrained to the same register (e.g. at calls). We fix this by ensuring the minimal bundle is trimmed only to the half of the instruction that overlaps the original LR.

This is analogous to the third fix in #74, but on the other end (start of LR rather than end of it).

### DIFF
--- a/src/ion/process.rs
+++ b/src/ion/process.rs
@@ -800,6 +800,7 @@ impl<'a, F: Function> Env<'a, F> {
         for entry_idx in 0..self.bundles[bundle.index()].ranges.len() {
             // Iterate manually; don't borrow `self`.
             let entry = self.bundles[bundle.index()].ranges[entry_idx];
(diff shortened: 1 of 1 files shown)


cfallin:
This fixes a fuzzbug (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51699 for those with access) and thus a compile panic on `main`; hopefully we can get it merged soon :-) Happy to talk through the fuzz case if anyone would like that.

jameysharp (approved):
Hey, I actually understood this one!

PR #89, opened 2022-09-25
Fix documentation for `inst_clobbers`

Amanieu:
Clobbers don't conflict with early uses, so they effectively act as a late def, not an early def.


cfallin (approved):
Thanks!

PR #95, opened 2022-10-07
Some small perf improvements

jameysharp:
According the Sightglass, these changes collectively are worth a 3% speedup by instructions retired, and 1-2% speedup by CPU cycles, when using Wasmtime to compile the bz2 and pulldown-cmark benchmarks. Hyperfine also reports a 2% speedup for each of the bz2, pulldown-cmark, and spidermonkey benchmarks.

### DIFF
--- a/src/indexset.rs
+++ b/src/indexset.rs
@@ -36,73 +36,41 @@ impl AdaptiveMap {
         }
     }
 
-    /// Expand into `Large` mode if we are at capacity and have no
-    /// zero-value pairs that can be trimmed.
-    #[inline(never)]
-    fn expand(&mut self) {
-        match self {
-            &mut Self::Small {
-                ref mut len,
-                ref mut keys,
-                ref mut values,
-            } => {
-                // Note: we *may* remain as `Small` if there are any
-                // zero elements. Try removing them first, before we
-                // commit to a memory allocation.
-                if values.iter().any(|v| *v == 0) {
-                    let mut out = 0;
-                    for i in 0..(*len as usize) {
-                        if values[i] == 0 {
-                            continue;
-                        }
-                        if out < i {
-                            keys[out] = keys[i];
-                            values[out] = values[i];
-                        }
-                        out += 1;
-                    }
-                    *len = out as u32;
-                } else {
-                    let mut map = FxHashMap::default();
-                    for i in 0..(*len as usize) {
-                        map.insert(keys[i], values[i]);
-                    }
-                    *self = Self::Large(map);
-                }
-            }
-            _ => {}
-        }
-    }
     #[inline(always)]
     fn get_or_insert<'a>(&'a mut self, key: u32) -> &'a mut u64 {
         // Check whether the key is present and we are in small mode;
         // if no to both, we need to expand first.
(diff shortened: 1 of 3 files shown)


cfallin (approved):
All looks reasonable, thanks!

The inlining of `AdaptiveMap::expand()` was I guess to avoid the duplicate `match self` discriminant check?

jameysharp:
> The inlining of `AdaptiveMap::expand()` was I guess to avoid the duplicate `match self` discriminant check?

That was a nice bonus, but not the goal. I tried to document the details in the commit message but GitHub review makes it hard to notice those. I hope that looking at d947e0154b8febfe4f004b7c69cf2d20425081cd helps?

Also, I meant to note that 869eef68608116b1c5e7522a0bbb3dea885883fc didn't have any effect on performance (and wasn't supposed to), but was setup for a change I tried that didn't have enough impact to seem worth doing. Nothing else in the PR depends on it. So would you take a look at that commit and see whether you think it's individually worth merging? I think the only thing to evaluate it on is code-style preference.

cfallin:
> That was a nice bonus, but not the goal. I tried to document the details in the commit message but GitHub review makes it hard to notice those. I hope that looking at https://github.com/bytecodealliance/regalloc2/commit/d947e0154b8febfe4f004b7c69cf2d20425081cd helps?

Ah, that makes more sense -- thanks! That change makes a lot more sense given the context.

> Also, I meant to note that https://github.com/bytecodealliance/regalloc2/commit/869eef68608116b1c5e7522a0bbb3dea885883fc didn't have any effect on performance (and wasn't supposed to), but was setup for a change I tried that didn't have enough impact to seem worth doing. Nothing else in the PR depends on it. So would you take a look at that commit and see whether you think it's individually worth merging? I think the only thing to evaluate it on is code-style preference.

Hmm, yeah, actually with no other reasons to do so, I prefer to keep the bare struct fields accessible (the ctor and accessors are just unnecessary code). If it's down to preference I'd prefer to drop that commit, but I don't care too much either way!

jameysharp:
> Hmm, yeah, actually with no other reasons to do so, I prefer to keep the bare struct fields accessible (the ctor and accessors are just unnecessary code). If it's down to preference I'd prefer to drop that commit, but I don't care too much either way!

I was leaning that way too. I'm not sure why I didn't just drop it to begin with...

For the record, the change that was setup for was to store the 64-bit key, pre-packed, in `PosWithPrio`, so that `PosWithPrio::key` does no computation at all. It currently compiles to three instructions, plus `ret`. It's called 2.5 million times when compiling Spidermonkey, so those three instructions add up, a little. To do that, I needed to hide the implementation of the field accessors, so they could unpack the key correctly.

I think the impact was something like a 0.1% improvement in compile time, which is impressive considering how few instructions it affected, but didn't seem like enough to justify bothering. Just inlining the key function made a much bigger difference.

The comments on e.g. `u64_key` suggest that at some point it compiled to a no-op if the struct layout worked out right, but I didn't observe that to be true now. Maybe it is again with `PosWithPrio::key` being inlined now; I haven't checked what code that generated at the call sites.

PR #96, opened 2022-10-21
Add explicit dependency on `libfuzzer-sys`

alexcrichton:
This fixes a build issue with the most recent version of `libfuzzer-sys` where the prior version of the `fuzz_target!` macro didn't need the crate as an explicit dependency but the latest version does.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -13,6 +13,7 @@ cargo-fuzz = true
 regalloc2 = { path = "../", features = ["fuzzing"] }
 log = { version = "0.4.8", default-features = false }
 env_logger = "0.8.3"
+libfuzzer-sys = "0.4.2"
 
 # Prevent this from interfering with workspaces
 [workspace]


alexcrichton:
The fuzz build also currently prints out:

```
warning: for loop over a `Result`. This is more readably written as an `if let` statement
  --> fuzz_targets/moves.rs:65:18
   |
65 |         for i in u.int_in_range(0..=2) {
   |                  ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(for_loops_over_fallibles)]` on by default
help: to check pattern in a loop use `while let`
   |
65 |         while let Ok(i) = u.int_in_range(0..=2) {
   |         ~~~~~~~~~~~~~ ~~~
help: consider using `if let` to clear intent
   |
65 |         if let Ok(i) = u.int_in_range(0..=2) {
   |         ~~~~~~~~~~ ~~~
```

which is just a warning so not fatal, but I wasn't sure what the intended code here was so I figured I'd ask.

jameysharp:
What libfuzzer-sys change caused this to be necessary? I'd expect this to be a bug in that crate, not something that should be worked around here. I think libfuzzer-sys 0.4.3 should have made this unnecessary, so I'd like to know what broke after that.

In #69 I removed this dependency because it requires keeping dependencies in sync across multiple `Cargo.toml` files. So I'd like to revert this PR but clearly we need something.

alexcrichton:
I'm not sure what the change was, I mostly just wanted to fix the fuzz build, I didn't investigate much futher.

PR #98, opened 2022-11-01
Avoid indexing and use iteration

fitzgen:
...while still appeasing the borrow checker by taking spillsets out of `self` and then putting them back in again when we're done.

I was just doing this to enable playing with some data structures in follow up commits, but decided to benchmark this commit as-is and found 2-4% speed ups to Cranelift compilation!

```
compilation :: instructions-retired :: benchmarks/bz2/benchmark.wasm

  Δ = 39946528.13 ± 38398.29 (confidence = 99%)

  no-index.so is 1.04x to 1.04x faster than main.so!

  [985704952 985984130.24 986180413] main.so
  [945649144 946037602.11 946262076] no-index.so

compilation :: instructions-retired :: benchmarks/pulldown-cmark/benchmark.wasm

  Δ = 48413802.56 ± 34288.05 (confidence = 99%)

  no-index.so is 1.03x to 1.03x faster than main.so!

  [1593663899 1593926801.92 1594246604] main.so
  [1545196678 1545512999.36 1545802144] no-index.so

compilation :: instructions-retired :: benchmarks/spidermonkey/benchmark.wasm

  Δ = 841028066.56 ± 253404.59 (confidence = 99%)

  no-index.so is 1.02x to 1.02x faster than main.so!

  [34798712681 34801346430.28 34802786661] main.so
  [33958847844 33960318363.72 33962177143] no-index.so
```

### DIFF
--- a/src/ion/spill.rs
+++ b/src/ion/spill.rs
@@ -89,9 +89,11 @@ impl<'a, F: Function> Env<'a, F> {
         spillslot: SpillSlotIndex,
     ) {
         self.spillsets[spillset.index()].slot = spillslot;
-        for i in 0..self.spillsets[spillset.index()].vregs.len() {
-            // don't borrow self
-            let vreg = self.spillsets[spillset.index()].vregs[i];
+
+        // Take `spillsets` to avoid a conflicting borrow of `self`.
+        let spillsets = std::mem::take(&mut self.spillsets);
+
+        for vreg in &spillsets[spillset.index()].vregs {
             trace!(
                 "spillslot {:?} alloc'ed to spillset {:?}: vreg {:?}",
                 spillslot,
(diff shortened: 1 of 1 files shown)


cfallin (approved):
Nice, thanks!

I recall a measurement by @Amanieu a while back that showed removing all indexing bounds-checks got something like 8% speedup. At the time we decided not to pursue this but that was because the approach on the table was unsafe indexing, not"appease borrowck with more clever means" as we see here. Given that, I suspect there may be other places where this matters too.

cfallin:
([Here](https://github.com/bytecodealliance/regalloc2/pull/25#issuecomment-1013111742) is the investigation I recalled from Amanieu)

jameysharp (commented):
Where is the conflicting borrow if you directly iterate over `self.spillsets`? I don't see any methods being invoked on `self`, and otherwise I've found Rust is smart enough to treat borrows of different fields as non-overlapping.

fitzgen:
> Where is the conflicting borrow if you directly iterate over `self.spillsets`? I don't see any methods being invoked on `self`, and otherwise I've found Rust is smart enough to treat borrows of different fields as non-overlapping.

Pushed a new commit to https://github.com/bytecodealliance/regalloc2/pull/99 to resolve this. I guess this is vestigial!

PR #100, opened 2022-11-02
Use maximum inline capacity available for `SmallVec<VRegIndex>` in `SpillSet`

fitzgen:
We were using 2, which is the maximum for 32-bit architectures, but on 64-bit architectures we can get 4 inline elements without growing the size of the `SmallVec`.

This is a statistically significant speed up, but is so small that our formatting of floats truncates it (so less than 1%).

```
compilation :: instructions-retired :: benchmarks/bz2/benchmark.wasm

  Δ = 3360297.85 ± 40136.18 (confidence = 99%)

  more-inline-capacity.so is 1.00x to 1.00x faster than main.so!

  [945563401 945906690.73 946043245] main.so
  [942192473 942546392.88 942729104] more-inline-capacity.so

compilation :: instructions-retired :: benchmarks/pulldown-cmark/benchmark.wasm

  Δ = 1780540.13 ± 39362.84 (confidence = 99%)

  more-inline-capacity.so is 1.00x to 1.00x faster than main.so!

  [1544990595 1545359408.41 1545626251] main.so
  [1543269057 1543578868.28 1543851201] more-inline-capacity.so

compilation :: instructions-retired :: benchmarks/spidermonkey/benchmark.wasm

  Δ = 36577153.54 ± 243753.54 (confidence = 99%)

  more-inline-capacity.so is 1.00x to 1.00x faster than main.so!

  [33956158997 33957780594.50 33959538220] main.so
  [33919762415 33921203440.96 33923023358] more-inline-capacity.so
```

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -264,9 +264,22 @@ pub struct BundleProperties {
     pub fixed: bool,
 }
 
+#[cfg(target_pointer_width = "64")]
+pub type VRegIndexList = SmallVec<[VRegIndex; 4]>;
+#[cfg(target_pointer_width = "32")]
+pub type VRegIndexList = SmallVec<[VRegIndex; 2]>;
+
+#[test]
+fn vreg_index_list_is_same_size_as_vec() {
+    assert_eq!(
+        std::mem::size_of::<VRegIndexList>(),
+        std::mem::size_of::<Vec<VRegIndex>>()
+    );
+}
+
 #[derive(Clone, Debug)]
 pub struct SpillSet {
-    pub vregs: SmallVec<[VRegIndex; 2]>,
+    pub vregs: VRegIndexList,
     pub slot: SpillSlotIndex,
     pub reg_hint: PReg,
     pub class: RegClass,
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -264,22 +264,26 @@ pub struct BundleProperties {
     pub fixed: bool,
 }
 
-#[cfg(target_pointer_width = "64")]
-pub type VRegIndexList = SmallVec<[VRegIndex; 4]>;
-#[cfg(target_pointer_width = "32")]
-pub type VRegIndexList = SmallVec<[VRegIndex; 2]>;
-
-#[test]
-fn vreg_index_list_is_same_size_as_vec() {
-    assert_eq!(
-        std::mem::size_of::<VRegIndexList>(),
-        std::mem::size_of::<Vec<VRegIndex>>()
-    );
+/// Calculate the maximum `N` inline capacity for a `SmallVec<[T; N]>` we can
+/// have without bloating its size to be larger than a `Vec<T>`.
+const fn no_bloat_capacity<T>() -> usize {
+    // `Vec<T>` is three words: `(pointer, capacity, length)`.
+    //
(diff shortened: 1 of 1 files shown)


src/ion/data_structures.rs
```diff
@@ -264,9 +264,22 @@ pub struct BundleProperties {
...
+#[cfg(target_pointer_width = "32")]
+pub type VRegIndexList = SmallVec<[VRegIndex; 2]>;
+
+#[test]
+fn vreg_index_list_is_same_size_as_vec() {
+    assert_eq!(
+        std::mem::size_of::<VRegIndexList>(),
+        std::mem::size_of::<Vec<VRegIndex>>()
+    );
+}
```
jameysharp, line 278: In other projects I've made use of the facts that

- `std::mem::size_of` is a `const fn` and
- array lengths can be arbitrary constant expressions

to make this smaller:

```suggestion
use std::mem::size_of;
pub type VRegIndexList = SmallVec<[VRegIndex; size_of::<usize>() * 2 / size_of::<VRegIndex>()]>;
```

This pattern does assume the element type is smaller than 2 `usize`; otherwise the result is just a slower version of `Vec`.

Sometimes I think it would be nice if the `smallvec` crate had a helper specifically for this pattern.

jameysharp (approved):
Even though the effect size is tiny, I think it's worth doing this. I do have a suggestion for a shorter way to write it though.

src/ion/data_structures.rs
cfallin: Can we turn this into a `const fn` helper?

jameysharp: What I'd really like is something like this:

```rust
type SmallestVec<T> = SmallVec<[T; size_of::<usize>() * 2 / size_of::<T>()]>;
```

Unfortunately, rustc says "type parameters may not be used in const expressions".

But yes, based on a quick Rust Playground experiment, it looks like this should work:

```rust
const fn elems<T>() -> usize {
    use std::mem::size_of;
    size_of::<usize>() * 2 / size_of::<T>()
}
pub type VRegIndexList = SmallVec<[VRegIndex; elems::<VRegIndex>()]>;
```

cfallin: Cool; something like that, with the helper named in a self-documenting way (`max_elems_in_smallest_smallvec` or somesuch) seems good to me.

fitzgen: I tried something similar initially but unfortunately because division can panic, it isn't `const` safe.

fitzgen: I can make the capacity a `const` function, but it would just be something like

```rust
const fn elems() -> usize {
    if cfg!(target_pointer_width = "64") {
        4
    } else {
        2
    }
}
pub type VRegIndexList = SmallVec<[VRegIndex; elems()]>;
```

because of the lack of division availability.

fitzgen: and we would still want to have the assertion in that case.

jameysharp: I'm not sure what you tried, but division isn't a problem: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=71cb6f52cf7300902a4c17390fc42f95

fitzgen: Weird... I don't know what was happening.

jameysharp (approved):
Looks great, and the comments help a lot.

PR #102, opened 2022-11-29
Expose ssa verification as a regalloc2 option

elliottt:
Adds the `validate_ssa` flag to the `RegallocOptions` struct, enabling ssa validation of inputs before register allocation takes place.


cfallin (approved):
LGTM, but would you mind doing the version bump as a separate PR, just for clarity / to keep the release process uniform?

PR #103, opened 2022-11-29
Bump the checkout action to v3 to avoid node12 deprecation

elliottt:
Recent builds show deprecation warnings for `node12` coming from our use of the `checkout@v2` action. This PR bumps the uses of the checkout action to v3 to avoid the deprecation.


cfallin (approved):
Thanks!

PR #105, opened 2022-11-30
SSA Validator Fixes

elliottt:
- Don't check for defs on a use of a fixed-nonallocatable operand
- Add traces to the ssa validator to aid debugging


cfallin (approved):
Good catch, thanks!

PR #108, opened 2023-01-24
Remove pinned VRegs.

cfallin:
This PR removes "pinned vregs" from regalloc2.

A pinned vreg is a vreg that is constrained to always be in a particular physical register at every operand that mentions it. (The value may be moved between its def/uses, but this should be transparent.)

Pinned vregs were built as a sort of compatibility feature for regalloc.rs' interface: that allocator allowed the embedder to mention physical registers directly in operands, and it would track liveranges on physical registers just like virtual registers.

In contrast, regalloc2 operands purely on vregs, and can track liveranges only on vregs. In order to emulate the above functionality, we set aside special vregs in order to stand in for the pregs' allocation state, and generated the implied constraints at every mention.

Pinned vregs were especially complex (and the source of quite a few fuzzbugs) because of their interaction with program move instructions, and the work to make it all efficient. We had four cases for moves (!): vreg->vreg, vreg->pinned, pinned->vreg, and pinned->pinned. Each was its own case with its own special logic. This all interacted poorly with the "no overlaps" invariant as well: we had to punch holes in liveranges and do other ugly things to generate liveranges with constraints that efficiently encoded the problem.

I've wanted to delete this code ever since writing it; and now that @elliottt has done the work on the Cranelift side to migrate all uses of pinned vregs over to ordinary vreg operands with register constraints, we can finally do that. This PR *should* be a pure "constant propagation and branch folding" action: make `is_pinned_vreg()` always return `None`, propagate that through to delete a bunch of conditional code, then remove that method altogether. (This implies a semver bump in the next release because of that method removal in the `Function` trait.)

Should be no-functional-change as long as the embedder isn't using pinned vregs!

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -366,25 +366,8 @@ impl CheckerState {
         }
     }
 
-    fn initial_with_pinned_vregs<F: Function>(f: &F) -> CheckerState {
-        // Scan the function, looking for all vregs that are pinned
-        // vregs, gathering them with their PRegs.
-        let mut pinned_vregs: FxHashMap<VReg, PReg> = FxHashMap::default();
-        visit_all_vregs(f, |vreg: VReg| {
-            if let Some(preg) = f.is_pinned_vreg(vreg) {
-                pinned_vregs.insert(vreg, preg);
-            }
-        });
-
-        let mut allocs = FxHashMap::default();
-        for (vreg, preg) in pinned_vregs {
-            allocs.insert(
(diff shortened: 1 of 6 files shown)


cfallin:
(This is less painful to review with "hide whitespace" enabled due to some un-indenting.)

Amanieu (approved):

elliottt (approved):
This looks great! :tada:

PR #109, opened 2023-01-24
Remove mod operands.

cfallin:
(Stacked on top of #108; only the second commit is the real change here.)

This PR removes "mod" or "modify" operands. regalloc2 currently supports non-SSA input -- that is, code with multiple defs for the same register -- and as part of this support, a kind of operand that both uses the old value of a vreg and redefines it with a new value.

This operand kind was included for compatibility with regalloc.rs, because quite a few instructions in Cranelift used it to naturally describe the machine-level semantics. (For example, on x86, most ALU instructions modify one operand to replace it with the result.)

However, we've been pushing to refactor toward pure-SSA input, and thanks to @elliottt's work in Cranelift, non-SSA support is no longer needed in regalloc2. Instead, instructions now use "reused-input" operand constraints: a new vreg is defined with the result, and constrained to be in the same allocation as the input vreg. The "mod operand" thus was split into two halves.

Since we now want to make use of the pure-SSA properties for other work (e.g., better splitting and simpler liverange-construction frontend by allowing overlapping LRs for one vreg), we should remove features that imply non-SSA input, including "mod" operands.

The main part of this PR should be a fairly straightforward mechanical change: it removes the `OperandKind` then deletes the cases that handle it.

In addition, since the `OperandKind` now needs only 1 bit instead of 2 in the 32-bit bitpacked `Operand` encoding, there was an opportunity to allocate this bit to some other purpose. I've opted to put it in the `RegClass` field, because there has been a latent need to define more than two register classes for some architectures.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -426,14 +426,13 @@ impl std::fmt::Display for OperandConstraint {
     }
 }
 
-/// The "kind" of the operand: whether it reads a vreg (Use), writes a
-/// vreg (Def), or reads and then writes (Mod, for "modify").
+/// The "kind" of the operand: whether it reads a vreg (Use) orwrites
+/// a vreg (Def).
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub enum OperandKind {
     Def = 0,
-    Mod = 1,
-    Use = 2,
+    Use = 1,
 }
 
 /// The "position" of the operand: where it has its read/write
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -405,14 +405,13 @@ impl std::fmt::Display for OperandConstraint {
     }
 }
 
-/// The "kind" of the operand: whether it reads a vreg (Use), writes a
-/// vreg (Def), or reads and then writes (Mod, for "modify").
+/// The "kind" of the operand: whether it reads a vreg (Use) orwrites
```
Amanieu, line 408: "orwrites"

Amanieu (approved):

Amanieu:
I presume there is also another follow up PR to remove program moves, which are no longer needed with SSA?

cfallin:
> I presume there is also another follow up PR to remove program moves, which are no longer needed with SSA?

I actually hadn't gotten that far yet, and wanted to verify that they weren't used anywhere performance-critical first. In principle a program move could still be used in SSA-pure code (a trivial unary op that defs a new vreg), but indeed many of the major reasons to use them (x86 mod operands etc) have gone away.

src/lib.rs
cfallin: Thanks!

PR #116, opened 2023-02-28
Misc refactorings

elliottt:
A few refactorings I made while reading through the regalloc2 source. The diff looks larger than it is because of indent changes, so I'd recommend reviewing with space changes turned off.

- Use a while-let instead of checking is_empty and popping
- This conditional should always be true, as we expect the input is in ssa
- Use iter_mut instead of iterating the index
- We don't support multiple defs of the same vreg anymore
- Drain instead of clear


cfallin (approved):
LGTM, thanks!

Issue #91, opened 2022-09-25
Support branch instructions with operands


Amanieu:
Currently branch instructions that take blockparams (which implies that the target block has multiple predecessors) are not allowed to have any operands. However on many targets a temporary register is needed to synthesize a jump whose target is outside the range of a single instruction.

For example on RISC-V any jump to a target outside a +/- 1MB range requires a `auipc tmp, %hi(target); jr tmp, %lo(target)` instruction sequence. I would like to be able to use an `EarlyDef` with a temporary vreg to tell the register allocator to select a free temporary register for `tmp`. Would it be possible to make this work?

cfallin:
This one's a bit tricky: the issue is the interference of "edge moves" with the uses/defs on the branch instruction. We need to put any moves that set up register state for the destination block prior to the branch itself; so conceptually, the branch sees the target block's register state.

We could require a stricter edge-splitting condition to work around this (currently we require no critical edges, but we do not require every edge to be split). However that would penalize the common case perhaps significantly.

What we do in Cranelift for cases like this that require a temp is to keep a temp register permanently set aside and not allocatable by the regalloc. This mostly comes into play when we need to generate spillslot offsets outside the range of load/store amode immediates.

(For the particular case of branch ranges we have the [MachBuffer](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/machinst/buffer.rs) that generates "branch veneers" in islands between blocks when branches go out of range; something like that may suit your use-case better?)

Amanieu:
> This one's a bit tricky: the issue is the interference of "edge moves" with the uses/defs on the branch instruction. We need to put any moves that set up register state for the destination block prior to the branch itself; so conceptually, the branch sees the target block's register state.
>
> We could require a stricter edge-splitting condition to work around this (currently we require no critical edges, but we do not require every edge to be split). However that would penalize the common case perhaps significantly.

I don't see how we could split the code any more than it already is. This situation happens with code of the form:

```
bb0:
	jump bb2(v0) // bb2 is out of range, need a scratch reg for the jump
bb1:
	jump bb2(v1) // bb2 is out of range, need a scratch reg for the jump
bb2(v2);
	...
```

Conceptually what I want is an unspillable vreg that starts in the jump instruction in bb0/bb1 and ends at the start of bb2. I think it might be possible to make this work, but more thought needs to be put into it.

[...]

PR #117, opened 2023-03-04
Re-introduce optional dedicated scratch registers

Amanieu:
Dedicated scratch registers used for resolving move cycles were removed in #51 and replaced with an algorithm to automatically allocate a scratch register as needed.

However in many cases, a client will already have a non-allocatable scratch register available for things like extended jumps (see #91). It makes sense to re-use this register for regalloc than potentially spilling an existing register.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1334,11 +1334,29 @@ pub struct MachineEnv {
     /// but still better than spilling.
     pub non_preferred_regs_by_class: [Vec<PReg>; 2],
 
+    /// Optional dedicated scratch register per class. This is needed to perform
+    /// moves between registers when cyclic move patterns occur. The
+    /// register should not be placed in either the preferred or
+    /// non-preferred list (i.e., it is not otherwise allocatable).
+    ///
+    /// Note that the register allocator will freely use this register
+    /// between instructions, but *within* the machine code generated
+    /// by a single (regalloc-level) instruction, the client is free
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1334,11 +1334,29 @@ pub struct MachineEnv {
...
+    /// necessary.
+    pub scratch_by_class: [Option<PReg>; 2],
+
     /// Some `PReg`s can be designated as locations on the stack rather than
     /// actual registers. These can be used to tell the register allocator about
     /// pre-defined stack slots used for function arguments and return values.
     ///
-    /// `PReg`s in this list cannot be used as an allocatable register.
+    /// `PReg`s in this list cannot be used as an allocatable or scratch
+    /// register.
```
cfallin, line 1359: Can we add a similar note above to the preferred and non-preferred allocatable sets, something to the effect of "if an explicit scratch register is provided in `scratch_by_class`, then it must not appear here"?

cfallin (approved):
Seems reasonable to have this option -- thanks! Just one request to add a clarifying doc-comment below.

cfallin:
@Amanieu if you'd like a new release with this feature, we can do that on Monday -- please feel free to send a PR to version-bump (this is a semver break so `0.7.0`) and we can merge it, ideally after #118 too so we take both API deltas at once.

PR #118, opened 2023-03-04
Remove support for program moves

### LINKED ISSUE #4: Remove support for non-SSA code
regalloc2 was initially designed for SSA inputs only (with blockparams rather than the more usual phi-nodes).

In the course of porting Cranelift to use regalloc2, we decided to build a shim in regalloc.rs on top of regalloc2, and to make this work, we added support for non-SSA code. This includes:

- Support for more than one def of a given vreg
- Support for "mod" (modify) operands that read and then write a vreg
- More efficient support for move instructions, which are used heavily by Cranelift to lower phis internally before passing code to regalloc

Once we adapt Cranelift's backends to pass SSA to regalloc2 directly instead of using the regalloc.rs API (which will be a significant effort on its own), we will be able to remove this functionality and simplify the code in various places, mostly in liverange computation and in move-generation. We should verify that we obtain equivalent code quality without the move coalescing in particular.

Amanieu:
This removes the last part of non-SSA support from regalloc2. Program moves allowed non-SSA code to copy a value from one vreg to another, but with SSA codegen this is not needed because the vreg itself can simply be reused directly.

Fixes #4

### DIFF
--- a/src/ion/moves.rs
+++ b/src/ion/moves.rs
@@ -179,8 +179,6 @@ impl<'a, F: Function> Env<'a, F> {
 
         let mut blockparam_in_idx = 0;
         let mut blockparam_out_idx = 0;
-        let mut prog_move_src_idx = 0;
-        let mut prog_move_dst_idx = 0;
         for vreg in 0..self.vregs.len() {
             let vreg = VRegIndex::new(vreg);
             if !self.is_vreg_used(vreg) {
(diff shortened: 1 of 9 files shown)


src/ion/moves.rs
```diff
@@ -228,6 +226,7 @@ impl<'a, F: Function> Env<'a, F> {
                     );
                 }
 
+                // TODO: Is this even possible to trigger without prog-moves?
```
Amanieu, line 229: I'm not completely sure about this part. By my understanding, it shouldn't be possible to reach these conditions without progmoves but I'm not 100% sure.

cfallin: This condition is triggered whenever there is a split, actually -- e.g., if we're forced to move a value to a different register prior to an instruction because another higher-priority liverange def'd by that instruction gets that register, then the split could happen at an arbitrary point within the block.

(FWIW, the question could also be answered by fuzzing -- I suspect if you add a panic here you'll see it hit shortly, even given that fuzzing has never generated functions with program-moves.)

cfallin (approved):
I love to see so much code get removed -- thank you for doing this! @elliottt and I were just talking about removing it now that we think we've removed all of the uses  in Cranelift.

A TODO comment added below can be removed I think (re: adjacent LRs in one block) but otherwise this looks like a straightforward branch-folding (`is_move()` -> always `None`) and DCE to me.

Amanieu:
Removed the TODO. You're right, it did trigger in the fuzzer.

PR #119, opened 2023-03-07
Make regalloc2 `#![no_std]`

Amanieu:
This crate doesn't require any features from the standard library, so it can be made `no_std` to allow it to be used in environments that can't use the Rust standard library.

This PR mainly performs the following mechanical changes:
- `std::collections` is replaced with `alloc::collections`.
- `std::*` is replaced with `core::*`.
- `Vec`, `vec!`, `format!` and `ToString` are imported when needed since they are no longer in the prelude.
- `HashSet` and `HashMap` are taken from the `hashbrown` crate, which is the same implementation that the standard library uses.
- `FxHashSet` and `FxHashMap` are typedefs in `lib.rs` that are based on the `hashbrown` types.

The only functional change is that `RegAllocError` no longer implements the `Error` trait since that is not available in `core`.

Dependencies were adjusted to not require `std` and this is tested in CI by building against the `thumbv6m-none-eabi` target that doesn't have `std`.


cfallin:
This mostly doesn't look too bad, so I'm happy to accommodate. One request:

> The only functional change is that RegAllocError no longer implements the Error trait since that is not available in core.

Is there a way to put this under a feature (e.g. `std` or `std-error`)?

cfallin (approved):

PR #120, opened 2023-04-02
Slightly more efficient vec initialization

Johan-Mi:
This will, at least on x86_64, compile down to simpler, shorter assembly that uses a zeroed allocation instead of a regular allocation, a memset and various `raw_vec` methods.


bnjbvr (approved):
Indeed, thanks!

cfallin (approved):

PR #121, opened 2023-04-04
Auto-fix Clippy lints

Johan-Mi:
I found no information about the MSRV of this library, so I just told Clippy to use 1.63.0, 5 versions older than the current stable.

This brings down Clippy's warning count from 66 to 25.

### DIFF
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -108,7 +108,7 @@ impl CFGInfo {
 
             approx_loop_depth.push(cur_depth);
 
-            while backedge_stack.len() > 0 && backedge_out[block] > 0 {
+            while !backedge_stack.is_empty() && backedge_out[block] > 0 {
                 backedge_out[block] -= 1;
                 *backedge_stack.last_mut().unwrap() -= 1;
                 if *backedge_stack.last().unwrap() == 0 {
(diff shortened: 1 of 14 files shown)


cfallin:
Hi @Johan-Mi -- unfortunately I have somewhat mixed feelings about Clippy in general, and in particular, it sometimes produces "improvements" that actually make code worse. A good example is [this](https://github.com/bytecodealliance/regalloc2/pull/121/files#diff-9e6512301206c0e64181f6230691d7f611a24201a875b3dc69355a67d6083201R114) part of your diff: it moved a call to a method that has a side effect, nominally the "main part" of that stanza of code, into the tail of an if-condition and makes it easy to miss.

I could quibble with some of the smaller lints too: for example, I personally prefer that `v.len() > 0` is more clear to read sometimes than `!v.is_empty()`; the latter spreads the description of the property to be queried across the two ends of the expression (the `!` at the beginning and `is_empty` at the end) and makes it easy to mis-read as the opposite condition.

So in general, I see it as fairly over-opinionated, and I think taking the human judgment out of the loop on how to best express an idea with idioms is a recipe for hard-to-understand code.

For these sorts of reasons, we don't run4 Clippy lints in this project, and I'm skeptical of "auto-fix" diffs. If you have specific things you want to suggest as part of a (human-authored) PR, I'm happy to talk about them, of course.

PR #123, opened 2023-04-06
Avoid using Vec::clear in clear_alloc

elliottt:
Iterate over the result of `Vec::drain` instead of iterating and then explicitly clearing the vector. Split out of #122 to minimize the diff.


cfallin (approved):
LGTM!

Issue #47, opened 2022-05-16
Support more than two register classes


yuyang-ok:
Riscv architecture have three category of registers.
integer registers are "x0-x31",
float     registers are  "f0-f31".
vector  registers are  "v0-v31".
riscv registers are not so general purpose as other platform.

cfallin:
Thanks @yuyang-ok -- we can possibly do this, but the issue is that we have a very tight bit-packing in `Operand` that leaves only one bit for the class.

We could add another bit (allowing four classes), but at the cost of the vreg index; this would limit the maximum vreg count to 1M rather than 2M. I'd rather try to avoid that if we can.

Another bit we could steal is in the "kind" field: we currently have two bits to support `Use`, `Def`, `Mod`. If we can eliminate all of the `Mod`s in Cranelift's use of regalloc2, then we could shrink back to use/def only.

In the meantime, you could temporarily solve the problem by coarsening the allocation. In other words, treat float and vector registers as one "register" allocated all at once. If the regalloc gives you "float/vec 15", then you can use both f15 and v15. It is wasteful, but temporarily solves the problem until we can do better. Does that seem workable?

yuyang-ok:
I can wait a little while,still have a lot work to do.

yuyang-ok:
@cfallin are you working on remove and `Mod` and add `Vector` class??

cfallin:
@yuyang-ok yes, we're actually making good progress toward that. I've merged a few PRs in Cranelift recently removing most uses of `mod` operands; there are just a few left in s390x.

yuyang-ok:
are you modifing `regalloc2`?? maybe I can help a little to add `Vector` class register.

cfallin:
Not yet -- we need to reclaim the bit in the 32-bit encoding of operands first, before we add another regclass. Otherwise we shrink the number of vregs that can be supported by half (from 2M to 1M), and this is starting to get small enough that people with very large inputs can hit it (e.g. bytecodealliance/wasmtime#4865).

yuyang-ok:
@cfallin ok.

PR #126, opened 2023-04-26
Add a third register class

afonso360:
👋 Hey,

For machines with completely separate vector registers it is useful to have a separate third register class. It looks like we now have encoding space to support that.

I've run all 5 fuzzers for about a couple of hours, and nothing showed up.

Fixes: #47

### DIFF
--- a/src/fuzzing/func.rs
+++ b/src/fuzzing/func.rs
@@ -146,7 +146,7 @@ impl Function for Func {
     fn spillslot_size(&self, regclass: RegClass) -> usize {
         match regclass {
             RegClass::Int => 1,
-            RegClass::Float => 2,
+            RegClass::Float | RegClass::Vector => 2,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


cfallin (approved):
This looks great -- thanks!

Once this merges, I'm happy to r+ another PR to bump the version. I think this is a semver break because we added an enum item to `RegClass`, so that'd be `0.8.0`. I'm happy to `cargo publish` then and we can pull it into Cranelift.

(Incidentally, on the Cranelift side we'll need to bump [this constant](https://github.com/bytecodealliance/wasmtime/blob/57dabd3b0b952e7e08ad75d56d76d71afaf584ff/cranelift/codegen/src/machinst/reg.rs#L12) to allow representing Vec-class pregs in instructions. The verbiage there is a bit out-of-date as the preg-as-vreg index range isn't "pinned" per se, just an encoding; but we can refine in a separate CL PR!)

cfallin:
@afonso360 one thing I realized when discussing another regalloc-related question: the `PRegSet` is currently a bitset using one `u128`, so it can represent two classes with maximum 64 regs each (128 total) but with three classes we now need 192 bits. This would manifest as an error when computing clobbers as soon as you actually tried to use the new class. Would you mind extending it too (probably with another `u128` so we cover the eventual four possible classes with 256 bits)? Thanks!

PR #129, opened 2023-04-30
Fix several bugs

Amanieu:
This PR fixes 3 issues:
- `PReg::class` was broken by #126.
- `PRegSet` needs more than 128 bits for 3 register classes.
- `impl Display for Operand` was incorrect for fixed-non-allocatable operands.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,7 +125,7 @@ impl PReg {
     /// The register class.
     #[inline(always)]
     pub const fn class(self) -> RegClass {
-        match self.bits & (0b11 << Self::MAX_BITS) {
+        match (self.bits >> Self::MAX_BITS) & 0b11 {
             0 => RegClass::Int,
             1 => RegClass::Float,
             2 => RegClass::Vector,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -188,49 +188,54 @@ impl core::fmt::Display for PReg {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct PRegSet {
-    bits: u128,
+    bits: [u128; 2],
 }
 
 impl PRegSet {
     /// Create an empty set.
     pub const fn empty() -> Self {
-        Self { bits: 0 }
+        Self { bits: [0; 2] }
     }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -188,49 +188,54 @@ impl core::fmt::Display for PReg {
...
 
     /// Returns whether the given register is part of the set.
     pub fn contains(&self, reg: PReg) -> bool {
-        let bit = reg.index();
-        debug_assert!(bit < 128);
-        self.bits & 1u128 << bit != 0
+        debug_assert!(reg.index() < 256);
+        let bit = reg.index() & 127;
+        let index = reg.index() >> 7;
+        self.bits[index] & 1u128 << bit != 0
```
cfallin, line 205: This is correct wrt operator precedence but it causes me to pause for a moment -- for easier mental parsing, could we add parens around `1u128 << bit` ?

src/lib.rs
```diff
@@ -188,49 +188,54 @@ impl core::fmt::Display for PReg {
...
-        debug_assert!(bit < 128);
-        Self {
-            bits: self.bits | (1u128 << bit),
-        }
+        debug_assert!(reg.index() < 256);
+        let bit = reg.index() & 127;
+        let index = reg.index() >> 7;
+        let mut out = self;
+        out.bits[index] |= 1u128 << bit;
+        out
```
cfallin, line 215: Now that this logic has become slightly more complex, perhaps we could delegate to `add` with something like

```
let mut out = self;
out.add(reg);
out
```
?

cfallin (approved):
Thanks! Just a few requests below but basically looks good.

src/lib.rs
Amanieu: I tried that at first, but it's not possible since `with` is a `const fn`. `add` can't be made `const` because it uses `&mut self`.

src/lib.rs
```diff
@@ -125,7 +125,7 @@ impl PReg {
     /// The register class.
     #[inline(always)]
     pub const fn class(self) -> RegClass {
-        match self.bits & (0b11 << Self::MAX_BITS) {
+        match (self.bits >> Self::MAX_BITS) & 0b11 {
```
cfallin, line 128: And, yikes, thanks for catching this -- sorry I missed it in review!

src/lib.rs
Amanieu: Done.

cfallin (approved):
Thanks!

PR #132, opened 2023-05-08
Panic when spillslot is too large

afonso360:
👋 Hey,

When a spill slot is larger than 256, we were silently truncating the top bits. This would cause the wrong spill size to silently be accepted. Instead panic when we can't support spills of that size.

### DIFF
--- a/src/ion/merge.rs
+++ b/src/ion/merge.rs
@@ -19,6 +19,7 @@ use crate::{
     ion::data_structures::BlockparamOut, Function, Inst, OperandConstraint, OperandKind, PReg,
 };
 use alloc::format;
+use core::convert::TryFrom;
 use smallvec::smallvec;
 
 impl<'a, F: Function> Env<'a, F> {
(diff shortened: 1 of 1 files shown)


cfallin:
To add some context on the impact: Cranelift currently uses spillslot sizes only of 1 (Int) or 2 (Float) words, so won't hit this. The issue arises when considering large-vector support (which @afonso360 is currently working on). The two possible resolutions are either to reject (with an actual panic) spillslots of >= 256 words, or support them. The former means we can't go as high as we'd like in supporting RISC-V's maximum vector size, on the one hand; but on the other hand, IMHO if we are spilling more than 2KB on a single spill decision, we probably should reconsider things more broadly, as that has potential for big performance cliffs. So I'm personally in favor of catching this with e.g. `u8::try_from().unwrap()`, but others please let us know if there are cases we should consider!

(FWIW, the `u8` is because of struct-size efficiency tuning so we'd want to measure the impact on compile time if we increased it.)

elliottt:
The `size` field was removed from the `SpillSet` struct in 09371f0b7b5a51e7846b8a3f53969330b4aa1f60, so I'm going to close this PR.

PR #133, opened 2023-05-09
Separate vectors for each priority level

elliottt:
Instead of storing priority in each entry in `inserted_moves`, keep one vector for each priority level. This saves us four bytes per `inserted_moves` entry, and requires only that we sort on the program point of the moves, not the pair of program point and priority.

Profiling with both sightglass and hyperfine show that there is a very small improvement in compile time for benchmarks like `pulldown-cmark` and `bz2`, with larger benchmarks like `spidermonkey` showing the same performance. One downside is that the redundant move eliminator is less effective, as it's no longer working across priority levels. This comes up in the bz2 benchmark which sees a slight regression in runtime performance, due to excessive writes back to the stack of values that have spilled but are used in moves at different priority levels.

One option for tackling the unnecessary moves would be to sort each vector and then walk them in parallel, allowing the redundant move eliminator to see a more consistent state of the world when resolving moves. This would also allow us to avoid the final sort of edits, as we could construct the correct order up front.

Another option for managing the redundant moves introduced here would be to wait for #122 to land, and instead work towards removing the redundant move eliminator. Removing that part of move resolution will lead to an overall improvement 
in compile-time performance, and would unlock refactorings like this PR.

The diff is considerably easier to read with white space turned off, as the main loop of `resolve_inserted_moves` was indented by one level.

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -434,7 +434,7 @@ pub struct Env<'a, F: Function> {
     // was to the approprate PReg.
     pub multi_fixed_reg_fixups: Vec<MultiFixedRegFixup>,
 
-    pub inserted_moves: Vec<InsertedMove>,
+    pub inserted_moves: [Vec<InsertedMove>; InsertMovePrio::Count as usize],
 
     // Output:
     pub edits: Vec<(PosWithPrio, Edit)>,
@@ -636,13 +636,14 @@ impl LiveRangeSet {
 
 #[derive(Clone, Debug)]
 pub struct InsertedMove {
-    pub pos_prio: PosWithPrio,
+    pub pos: ProgPoint,
     pub from_alloc: Allocation,
     pub to_alloc: Allocation,
     pub to_vreg: VReg,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u8)]
 pub enum InsertMovePrio {
     InEdgeMoves,
     Regular,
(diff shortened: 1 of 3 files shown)


src/ion/data_structures.rs
```diff
@@ -417,7 +417,7 @@ pub struct Env<'a, F: Function> {
     // was to the approprate PReg.
     pub multi_fixed_reg_fixups: Vec<MultiFixedRegFixup>,
 
-    pub inserted_moves: Vec<InsertedMove>,
+    pub inserted_moves: [Vec<InsertedMove>; InsertMovePrio::Count as usize],
```
Amanieu, line 420: I like using the [enum-map](https://docs.rs/enum-map/) crate for this sort of thing, I find that it makes the code much cleaner.

bjorn3: enum-map uses a proc macro and unsafe code. Cranelift currently doesn't depend on any proc macros.

elliottt: That definitely looks like a nicer api than what I'm doing here! As depending on `enum-map` would require us to audit it as a dependency in cranelift, I'm inclined to stick with the current approach. Perhaps using a well-placed `debug_assert` would make the implementation cleaner?

src/ion/data_structures.rs
```diff
@@ -606,20 +606,37 @@ impl LiveRangeSet {
...
+#[repr(u8)]
 pub enum InsertMovePrio {
     InEdgeMoves,
     Regular,
     MultiFixedRegInitial,
     MultiFixedRegSecondary,
     ReusedInput,
     OutEdgeMoves,
+
+    /// The number of priorities, used to determine the length of the Env::inserted_moves array.
```
cfallin, line 625: Let's leave a comment here too noting that any new priorities should be added to `InsertMovePrio::all()` below.

src/ion/data_structures.rs
cfallin: @elliottt are you thinking an assert that the actual priority is < `Count`, or...? (That seems somewhat unnecessary to me but perhaps I'm not seeing somewhere else an assert would make sense)

Re: `enum-map`, I'll note that in addition to the concerns above, we try to be really careful about build time here (this is the reason why deps and proc macros in particular are avoided). I like that we have fairly lean crates low in the dependency graph and want to try to keep it that way :-)

cfallin (approved):
One note about a comment, otherwise LGTM.

elliottt:
After some fuzzing, it turns out that we can't yet make this refactoring: the redundant move eliminator may invalidate assumptions about register allocations at later priorities with no way to communicate those changes. We should reconsider this change once we've managed to remove the need for the redundant move eliminator.

PR #137, opened 2023-05-18
Prefer registers for block-param moves

elliottt:
Use a hash map for block-param move sources instead of a vector, using the source block and destination block-param as the key. These two values uniquely define the source of the edge, as we require critical edges to have been split prior to running RA2.

This change in block-param source storage allows us to skip sorting the move source and destination data structures, and gives us the opportunity to apply a heuristic when multiple sources for a block-param move exist. Currently we prefer register sources, which is the same heuristic we apply for inter-block moves as well.

### DIFF
--- a/src/ion/moves.rs
+++ b/src/ion/moves.rs
@@ -260,7 +260,11 @@ impl<'a, F: Function> Env<'a, F> {
             }
         }
 
-        let mut block_param_sources = FxHashMap::<BlockparamSourceKey, Allocation>::default();
+        let mut block_param_sources =
+            FxHashMap::<BlockparamSourceKey, Allocation>::with_capacity_and_hasher(
+                3 * self.func.num_insts(),
+                Default::default(),
+            );
         let mut block_param_dests = Vec::with_capacity(3 * self.func.num_insts());
 
         let debug_labels = self.func.debug_value_labels();
--- a/src/ion/moves.rs
+++ b/src/ion/moves.rs
@@ -17,7 +17,7 @@ use super::{
     RedundantMoveEliminator, VRegIndex, SLOT_NONE,
 };
 use crate::ion::data_structures::{
-    u128_key, u64_key, BlockparamIn, BlockparamOut, CodeRange, FixedRegFixupLevel, LiveRangeKey,
+    u64_key, BlockparamIn, BlockparamOut, CodeRange, FixedRegFixupLevel, LiveRangeKey,
     LiveRangeListEntry, PosWithPrio,
 };
 use crate::ion::reg_traversal::RegTraversalIter;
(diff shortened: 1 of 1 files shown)


elliottt:
Fuzzing overnight didn't find anything surprising :+1:

cfallin (approved):
This is a really fantastic simplification -- nice!

I'm curious, do you have any compile-time speedup numbers? (Fine to merge either way, it's clear it shouldn't be a slowdown, just curious...).

elliottt:
> I'm curious, do you have any compile-time speedup numbers? (Fine to merge either way, it's clear it shouldn't be a slowdown, just curious...).

Thanks for asking for these! It turned out that bz2 saw a slight slowdown without pre-allocating the hashmap for the blockparam move sources. With the preallocation it sees no slowdown, and pulldown-cmark sees a slight speed-up.

Here are the raw measurements:

## pulldown-cmark

```
compilation :: cycles :: benchmarks/pulldown-cmark/benchmark.wasm                                                               

  Δ = 12393781.18 ± 5935062.46 (confidence = 99%)               

  587996ff63fe6776d91c43b072e466ad01bea523.so is 1.00x to 1.01x faster than main.so!                                            

  [2522572270 2540002385.66 2558628399] 587996ff63fe6776d91c43b072e466ad01bea523.so                                             
  [2537170438 2552396166.84 2568060398] main.so                 

instantiation :: cycles :: benchmarks/pulldown-cmark/benchmark.wasm                                                             

  No difference in performance.                                 

  [501323 547989.22 640962] 587996ff63fe6776d91c43b072e466ad01bea523.so                                                         
  [494232 543986.56 637545] main.so                             

execution :: cycles :: benchmarks/pulldown-cmark/benchmark.wasm                                                                 

  No difference in performance.                                 

  [23558532 23695690.16 23960060] 587996ff63fe6776d91c43b072e466ad01bea523.so                                                   
  [23542797 23674810.12 23886557] main.so 
```

## bz2

```
compilation :: cycles :: benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [1092328920 1111227909.30 1127657680] 587996ff63fe6776d91c43b072e466ad01bea523.so
  [1085276032 1105975193.78 1122258117] main.so

instantiation :: cycles :: benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [335560 383419.02 409518] 587996ff63fe6776d91c43b072e466ad01bea523.so
  [334540 383846.86 449780] main.so

execution :: cycles :: benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [303631417 306741908.28 311140560] 587996ff63fe6776d91c43b072e466ad01bea523.so
  [303443567 306799324.78 310570795] main.so
```

PR #138, opened 2023-05-24
Handle dedicated scratch registers in the redundant move eliminator

Amanieu:
When a dedicated scratch register is provided by the environment, the redundant move eliminator needs to assume that this register can be clobbered by any instruction.

This register can't be specified as a clobber because it is also used by jump instructions which cannot have operands.

### DIFF
--- a/src/ion/moves.rs
+++ b/src/ion/moves.rs
@@ -796,6 +796,13 @@ impl<'a, F: Function> Env<'a, F> {
                 for reg in this.func.inst_clobbers(inst) {
                     redundant_moves.clear_alloc(Allocation::reg(reg));
                 }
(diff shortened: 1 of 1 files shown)


Amanieu:
This wasn't caught by the fuzzer because our fuzzing setup doesn't use a dedicated scratch register. I could change that but then we wouldn't be testing the case where regalloc2 needs to find a scratch register on its own.

cfallin (approved):
Thanks! Do you need a point release for this?

Amanieu:
> Thanks! Do you need a point release for this?

That's not necessary since I'm currently using a fork that includes the changes from #131. Once that is resolved I'd like to switch back to using mainline regalloc2.

PR #140, opened 2023-06-01
Remove the redundant move eliminator

elliottt:
- Remove the redundant move eliminator
- Merge together spill ranges for the same vreg


Amanieu:
I'm getting a significant code size regression with this PR in my compiler:
```
   text	   data	    bss	    dec	    hex	filename
66919040	      0	      0	66919040	3fd1a80	before
69489712	      0	      0	69489712	4245430	after
```

elliottt:
> I'm getting a significant code size regression with this PR in my compiler:

Yes, that makes sense. I had been hoping that widening the ranges that were moved to the spill bundle would cut down on stack movement enough that we could completely remove the redundant move eliminator, but I'm noticing that there's still quite a lot of stack movement introduced in wasmtime as well.

I'm planning on continuing to investigate this problem, which is why the PR is still in a draft state. Once I've investigated other options for avoiding the additional stack movement, I'll either mark this ready or close it. Until then, I'm just using the PR as a good anchor for discussion.

elliottt:
I haven't had much time to work on this recently. I'll re-open the PR if I get some more time :)

PR #141, opened 2023-06-03
Remove the merged_into field of LiveRangeData

elliottt:
This field was never written to after initialization, and the only function that looked at its value was never called.


cfallin (approved):
This is left over from some old heuristic IIRC -- no use now; thanks for finding this!

PR #142, opened 2023-06-07
Add a newtypes for LiveRange, LiveBundle, SpillSet, and VReg storage

elliottt:
Wrap up the vectors used as storage in the `Env` type so that it's possible to define custom instances of `Index` and `IndexMut`. Additionally, move some creation operations onto those newtypes, instead of defining them on `Env`.


cfallin (commented):
Overall I like this a lot -- thanks for working on this!

I do wonder if it might reduce verbosity somewhat to use a macro that defines the wrapper struct, the index newtype, and the `Index`/`IndexMut` impls, and `len`/iter methods, all at once; something like `define_entity_index_and_vec!(LiveRangeIndex, LiveRange, LiveRanges);` (replacing the existing macro that defines just the index type). One can still add additional methods with another impl on the type; but this would give a uniform set of methods. What do you think?

elliottt:
I definitely considered having the macro produce the newtypes as well, I'll experiment with that tomorrow and see how it looks :)

src/ion/data_structures.rs
```diff
@@ -396,6 +396,70 @@ impl BlockparamIn {
...
+        self.storage.push(LiveRange {
+            range,
+            vreg: VRegIndex::invalid(),
+            bundle: LiveBundleIndex::invalid(),
+            uses_spill_weight_and_flags: 0,
+
+            uses: smallvec![],
+        });
+
+        LiveRangeIndex::new(idx)
```
cfallin, line 412: I wonder if we could make these `add` methods a little cleaner by generating a `push` method that returns the newtype'd index to the macro body, and using that? (So the body here would be just `self.push(LiveRange { ... })`)

src/ion/liveranges.rs
```diff
@@ -227,31 +192,26 @@ impl<'a, F: Function> Env<'a, F> {
...
 
-        if self.vregs[vreg.index()].ranges.is_empty()
+        if self.vregs[vreg].ranges.is_empty()
             || range.to
-                < self.ranges[self.vregs[vreg.index()]
-                    .ranges
-                    .last()
-                    .unwrap()
-                    .index
-                    .index()]
```
cfallin, line 237: This is *so* much nicer with less stuttering! "vreg vreg index ranges last unwrap index index" reads like some wild Forth sequence or something...

src/ion/liveranges.rs
```diff
@@ -797,20 +757,20 @@ impl<'a, F: Function> Env<'a, F> {
         // when needed, here and then again at the end of allocation
         // when resolving moves.
 
-        for vreg in &mut self.vregs {
+        for vreg in self.vregs.iter_mut() {
```
cfallin, line 760: I suppose we could implement `IntoIterator` to make this work again (and whatever the equivalent is for mut iters) but I don't mind too much if you want to punt on that either...

cfallin (approved):
Looks great; thanks so much! A few additional thoughts if you want to do some further cleanup but otherwise I'm happy to see this merge.

src/ion/data_structures.rs
elliottt: I like that idea! The fewer places the index value is directly manipulated, the better.

src/ion/liveranges.rs
elliottt: I added `IntoIterator` implementations mutable and non-mutable references, and reverted the places where I had switched to explicit use of `iter`/`iter_mut` :+1:

PR #143, opened 2023-06-08
Move edits out of Env

elliottt:
Return move edits from `resolve_inserted_moves` instead of eagerly allocating that vector in `Env`.

### DIFF
--- a/src/ion/mod.rs
+++ b/src/ion/mod.rs
@@ -135,11 +135,7 @@ pub fn run<F: Function>(
     }
 
     Ok(Output {
-        edits: edits
-            .edits
-            .into_iter()
-            .map(|(pos_prio, edit)| (pos_prio.pos, edit))
-            .collect(),
+        edits: edits.into_edits().collect(),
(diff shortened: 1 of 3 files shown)


src/ion/mod.rs
```diff
@@ -129,14 +128,14 @@ pub fn run<F: Function>(
...
-    env.run()?;
+    let edits = env.run()?;
 
     if enable_annotations {
         env.dump_results();
     }
 
     Ok(Output {
-        edits: env
+        edits: edits
```
cfallin, line 138: If we wanted to clean up just a bit more, we could add a method around this iterator chain on `Edits` (`into_...` or something). Happy to skip if you'd prefer not to though.

cfallin (approved):
LGTM, thanks!

src/ion/mod.rs
elliottt: I moved the iterator chain into a method that consumes `self`, and I think it cleaned things up nicely. Thanks for the suggestion!

PR #148, opened 2023-07-12
Add `regalloc2-tool`

Amanieu:
This PR has 2 components:
- A `SerializableFunction` which wraps a `Function` and `MachineEnv`. This type can be serialized and deserialized with Serde, and is enabled by the "enable-serde" feature.
- A `regalloc2-tool` binary which reads a bincode-encoded `SerializableFunction` and then runs the register allocator and checker on it.

This is a useful tool for debugging register allocation failures and to investigate cases of poor register allocation.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -10,8 +10,7 @@ targets = [
 vulnerability = "deny"
 unmaintained = "deny"
 yanked = "deny"
-ignore = [
-]
+ignore = []
 
 # https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
 [licenses]
@@ -19,6 +18,7 @@ allow = [
     "Apache-2.0 WITH LLVM-exception",
     "Apache-2.0",
     "MIT",
+    "Unicode-DFS-2016",
 ]
 
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
--- a/regalloc2-tool/src/main.rs
+++ b/regalloc2-tool/src/main.rs
@@ -1,0 +1,95 @@
+use std::path::PathBuf;
+
+use clap::Parser;
+use regalloc2::{
+    checker::Checker, serialize::SerializableFunction, Block, Edit, Function, InstOrEdit, Output,
+    RegallocOptions,
+};
+
+#[derive(Parser)]
+/// Tool for testing regalloc2.
+struct Args {
+    /// Print the input function and the result of register allocation.
+    #[clap(short = 'v')]
+    verbose: bool,
+
+    /// Input file containing a bincode-encoded SerializedFunction.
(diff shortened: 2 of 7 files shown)


regalloc2-tool/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    if args.verbose {
+        print_output(&function, &output);
+    }
+
+    let mut checker = Checker::new(&function, function.machine_env());
+    checker.prepare(&output);
+    if let Err(e) = checker.run() {
+        panic!("Regsiter allocation checker failed: {e:#?}");
```
cfallin, line 50: s/Regsiter/Register/

src/serialize.rs
```diff
@@ -0,0 +1,311 @@
...
+    debug_value_labels: Vec<(VReg, Inst, Inst, u32)>,
+    spillslot_size: Vec<usize>,
+    multi_spillslot_named_by_last_slot: bool,
+    allow_multiple_vreg_defs: bool,
+}
+
+impl SerializableFunction {
+    /// Creates a new `SerializableFunction` from an arbitray `Function` and
+    /// `MachineEnv`.
+    pub fn new(func: &impl Function, machine_env: MachineEnv) -> Self {
```
cfallin, line 60: Should the second arg be a `&MachineEnv`? Alternately, no clone two lines below?

deny.toml
```diff
@@ -10,15 +10,15 @@ targets = [
...
-]
+ignore = []
 
 # https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
 [licenses]
 allow = [
     "Apache-2.0 WITH LLVM-exception",
     "Apache-2.0",
     "MIT",
+    "Unicode-DFS-2016",
```
cfallin, line 21: Could we add a comment why this is needed?

Probably fine from first principles but any addition here does mean we need to update downstream users' allowlists as well (e.g. Cranelift) which is sometimes a little tricky / requires consultation to be sure.

If my guess that this comes from `pretty_env_logger` is correct, perhaps could we use just `env_logger`?

cfallin (approved):
Thanks a bunch -- this will be really useful to have!

deny.toml
Amanieu: It actually comes from `unicode-ident`, which comes from `syn`, which is itself used by `serde-derive`. I think this appears here because regalloc2-tool enables the `enable-serde` feature of the crate. It shouldn't affect any downstream users if they don't have `enable-serde` enabled.

src/serialize.rs
Amanieu: Removed the clone.

PR #149, opened 2023-07-12
Fix handling of dedicated scratch registers

Amanieu:
`get_reg` can be called multiple times if multiple scratch registers are needed (e.g. for stack-to-stack moves).


cfallin (approved):
Good catch -- thank you for finding this!

Tangential to regalloc2 itself but important information nonetheless: I just verified that Cranelift does not specify dedicated scratch registers on any of its four targets, so is not affected by this. The impact of this is limited to embedders that specify dedicated scratch registers.

PR #150, opened 2023-07-12
Check constraints for Def operands in the checker

Amanieu:
This was previously skipped: only Use operands were checked to ensure they had a valid allocation for their constraint.

No new fuzzbugs found, I just noticed this while looking over the checker.


cfallin (approved):
Ah, good catch, thanks!

PR #151, opened 2023-07-12
Avoid generating the state in the check if logging is disabled

Amanieu:
This significantly improves the performance of the checker when trace logging is not enabled.


cfallin (approved):
Thanks!

PR #152, opened 2023-07-13
Run the fuzzer with all register classes

Amanieu:
This quickly triggers a fuzzbug when two classes have the same spillslot size. It turns out that this is caused by the parallel move resolver not supporting cross-class moves, which can happen if the parallel involves a spillslot used by different classes before/after the move.

### DIFF
--- a/src/fuzzing/func.rs
+++ b/src/fuzzing/func.rs
@@ -30,18 +30,6 @@ pub struct InstData {
 }
 
 impl InstData {
-    pub fn op(def: usize, uses: &[usize]) -> InstData {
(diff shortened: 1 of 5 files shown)


Amanieu:
I found the cause of the fuzzbug, which should be fixed now.

cfallin (approved):
Good find, thank you!

This is I think not reachable from Cranelift because we have distinct spillslot sizes per class (Int == 8, Float == 16) but is definitely a real bug.

Amanieu:
It is reachable from Cranelift on RISC-V where both Int and Float use 1 spillslot.

cfallin:
Oof, you're right, thanks. We'll do a release!

PR #153, opened 2023-07-13
Version bump to 0.9.2.

cfallin:
This pulls in a fix for a stackslot confusion bug (among stack slots of the same size) from #152. The bug is reachable only if the embedder has the same stackslot size for different classes, and could result in sharing of a stackslot and thus moves between values of different class (which would likely cause panics in an embedder). This is reachable from RISC-V in Cranelift.


elliottt (approved):
:rocket:

PR #155, opened 2023-07-14
Allow merging bundles that have a fixed-reg def

Amanieu:
The previous check was overly conservative: we only need to reject bundles that have multiple uses in the same instruction.

This can only happen with an early use and a late def, so we round the start of each range containing a fixed def up to the start of its instruction to detect overlaps.

### DIFF
--- a/src/ion/merge.rs
+++ b/src/ion/merge.rs
@@ -14,7 +14,8 @@
 
 use super::{Env, LiveBundleIndex, SpillSet, SpillSlotIndex, VRegIndex};
 use crate::{
-    ion::data_structures::BlockparamOut, Function, Inst, OperandConstraint, OperandKind, PReg,
+    ion::data_structures::{BlockparamOut, CodeRange},
+    Function, Inst, OperandConstraint, OperandKind, PReg, ProgPoint,
 };
 use alloc::format;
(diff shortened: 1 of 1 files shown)


cfallin (commented):
Thanks a bunch -- as you probably figured, this code was "written in fuzzbug failures", but the `adjust_range_start` conservative overapproximation will *I think* handle the case I had discovered. If you haven't already, would you mind fuzzing with this continuously over a few days (I've seen issues take 12-24 hours to come up sometimes)? With that confirmed, I'm happy to merge this.

Amanieu:
At the very least this did fix the fuzzbug that I encountered when I simply removed the previous check. But sure, I'll run the fuzzer overnight to see if it hits anything.

Amanieu:
Fuzzer ran all day and night without issues. The new check is still somewhat conservative but at the very least it should be a net improvement over the current situation.

I believe the fully precise condition is:
- There is a late fixed def in the merged bundle.
- The merged bundle is live over the early part of the instruction with the fixed-def.
- There is an early fixed use at the same instruction as the fixed-def with the same preg *outside* the merged bundle.

cfallin (approved):
Great, thanks!

PR #156, opened 2023-08-01
Make parallel-moves resolver available publicly

jameysharp:
We want to also use this functionality in ABI glue in Cranelift.

I haven't tried using this module from Cranelift yet though. I'm mostly opening this PR as a quick check that there's nothing obviously wrong with just exposing the whole module.


cfallin (approved):
LGTM! It's useful functionality and there's no reason not to let crate users use it...

PR #157, opened 2023-08-04
Only sort parallel-moves list once

jameysharp:
Checking whether sources and dests overlap can use the same sort order that's needed later, as long as the check is updated to match.

I haven't done any benchmarking on this, but I have fuzzed it and didn't find any failures in 15 million inputs.

### DIFF
--- a/src/moves.rs
+++ b/src/moves.rs
@@ -44,11 +44,13 @@ impl<T: Clone + Copy + Default + PartialEq> ParallelMoves<T> {
     }
 
     fn sources_overlap_dests(&self) -> bool {
-        // Assumes `parallel_moves` has already been sorted in `resolve()` below.
-        for &(_, dst, _) in &self.parallel_moves {
+        // Assumes `parallel_moves` has already been sorted in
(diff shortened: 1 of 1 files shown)


src/moves.rs
```diff
@@ -44,11 +44,13 @@ impl<T: Clone + Copy + Default + PartialEq> ParallelMoves<T> {
     }
 
     fn sources_overlap_dests(&self) -> bool {
-        // Assumes `parallel_moves` has already been sorted in `resolve()` below.
-        for &(_, dst, _) in &self.parallel_moves {
+        // Assumes `parallel_moves` has already been sorted in
```
cfallin, line 47: can we clarify to "sorted by ..." for maximal explicitness here?

cfallin (approved):
LGTM, thanks!

src/moves.rs
jameysharp: Done, and rebased on top of #158 since you merged that. Like I contemplated in that PR but couldn't implement without also making this PR's changes, I've now also deferred checking overlaps until after deleting self-moves, since cycle resolution is not necessary for those.

PR #160, opened 2023-08-08
Revert "Allow merging bundles that have a fixed-reg def"

elliottt:
Reverts bytecodealliance/regalloc2#155

Fuzzing discovered a case that caused the panic on line [1241 of src/ion/process.rs](https://github.com/bytecodealliance/regalloc2/blob/52bf322832b57b6103a06df6d4df5c6d59f1540a/src/ion/process.rs#L1241) to trigger.

PR #161, opened 2023-08-08
Propagate the cached fixed-def flag when merging bundles

Amanieu:
This was missed in #155, and previously wasn't an issue since such
bundles were never merged.

This PR also un-reverts #160, since the underlying bug is fixed.

### DIFF
--- a/src/ion/merge.rs
+++ b/src/ion/merge.rs
@@ -152,6 +152,9 @@ impl<'a, F: Function> Env<'a, F> {
             if self.bundles[from].cached_fixed() {
                 self.bundles[to].set_cached_fixed();
             }
+            if self.bundles[from].cached_fixed_def() {
+                self.bundles[to].set_cached_fixed_def();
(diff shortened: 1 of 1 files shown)


cfallin:
@Amanieu I'll take a look at this and think in more detail about it in the next few days; in the meantime, would you be willing to fuzz this for a bit longer? It took oss-fuzz three weeks to find the prior bug. I don't think it's reasonable to ask a contributor to put that much testing into every patch (if you're willing, though, then fantastic!), but maybe at least 3-4 days would be good. Then subsequent to merging let's wait a bit, maybe two weeks or so, until we release (just as we do now for Cranelift).

Amanieu:
Fuzzing has been continuously running for the past 24 hours without issues. I doubt anything new will come up though considering how long it took to find the bug in the first place.

cfallin (approved):
LGTM, thanks!

PR #163, opened 2023-08-14
Remove unused function generator options

elliottt:
The `ssagen` and `ion_checker` fuzz targets are the only ones that consume the `Options` struct for function generation in a way that doesn't use the defaults. This PR removes all options that were the same across each use of `arbitrary_with_options`, and makes a small refactor to where we were collecting vregs to be defined in a block.

Co-authored-by: Jamey Sharp <jsharp@fastly.com>


cfallin (approved):
LGTM, thanks! Most of these were useful when bringing up RA2 initially but aren't as useful anymore so this makes sense.

PR #168, opened 2023-11-19
Allow fixed-early-def and fixed-late-use constraints

Amanieu:
These are useful to model fixed registers which should not be reused for other uses/defs. These were disallowed in #54, but this was too conservative.

Fundamentally, the only situation where a preg can be used in multiple fixed constraints within a single instruction is with an early-use and a late-def. Anything else is a user error because the live ranges will overlap.

As such this PR relaxes the operand rewrite from #54 to only apply in this specific situation. Fixed-late-use and fixed-early-def operands are left unchanged.

### DIFF
--- a/src/fuzzing/func.rs
+++ b/src/fuzzing/func.rs
@@ -471,27 +471,25 @@ impl Func {
                         let i = u.int_in_range(0..=(operands.len() - 1))?;
                         let op = operands[i];
                         let fixed_reg = PReg::new(u.int_in_range(0..=62)?, op.class());
+                        if op.kind() == OperandKind::Def && op.pos() == OperandPos::Early {
+                            // Early-defs with fixed constraints conflict with
+                            // any other fixed uses of the same preg.
+                            if fixed_late.contains(&fixed_reg) {
+                                break;
+                            }
+                        }
+                        if op.kind() == OperandKind::Use && op.pos() == OperandPos::Late {
+                            // Late-use with fixed constraints conflict with
+                            // any other fixed uses of the same preg.
+                            if fixed_early.contains(&fixed_reg) {
+                                break;
+                            }
+                        }
                         let fixed_list = match op.pos() {
                             OperandPos::Early => &mut fixed_early,
                             OperandPos::Late => &mut fixed_late,
                         };
                         if fixed_list.contains(&fixed_reg) {
-                            break;
-                        }
-                        if op.kind() != OperandKind::Def && op.pos() == OperandPos::Late {
-                            // Late-uses/mods with fixed constraints
-                            // can't be allowed if we're allowing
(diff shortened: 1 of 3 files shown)


elliottt (commented):
Could you explain the use-case for a late-use? I'm a little unclear about how that would be used, and also am not sure if we need to still worry about this comment with the changes in this pr:
```
                                // See note in fuzzing/func.rs: we
                                // can't allow this, because there
                                // would be no way to move a value
                                // into place for a late use *after*
                                // the early point (i.e. in the middle
                                // of the instruction).
```
Specifically, what guarantees that we'd never try to generate a move in the middle of an instruction to satisfy a late-use? Currently we unconditionally use the `CodeRange::from` program point when inserting moves, so if the liverange containing a late-use got trimmed to only include that use, what prevents us from trying to insert the move at the late position?

Amanieu:
I personally don't have a use case for fixed-late-use, I only needed fixed-early-def in my code. I just allowed fixed-late-use for symmetry.

> Specifically, what guarantees that we'd never try to generate a move in the middle of an instruction to satisfy a late-use? Currently we unconditionally use the `CodeRange::from` program point when inserting moves, so if the liverange containing a late-use got trimmed to only include that use, what prevents us from trying to insert the move at the late position?

Live range trimming/splitting will only split live ranges on instruction boundaries, never in the middle of an instruction. Since the value of a late use must necessarily come from a previous instruction, a move will simply be inserted before the instruction.

afonso360:
I was looking for the PR where I first needed this, and it looks like it was https://github.com/bytecodealliance/wasmtime/pull/6568 but weirdly, I added a fixed-late-use operand, and it worked out fine? It's still in there.

Here's some of the discussion regarding this: https://github.com/bytecodealliance/wasmtime/pull/6568#discussion_r1227347838

Now that I'm looking at this again I'm not sure if a fixed-late-use operand behaves exactly the same as a fixed-late use constraint?

---

As an explanation of what is going on in that PR, it introduces the `vslideup` instruction, which has some special constraints, namely:
* It modifies the destination register (`vd`); and
* None of the input registers (`vs2` or `vm`) must be the same register as the destination register (`vd`)
* The mask register (`vm`) is optional, but when provided is fixed to the `v0` register

elliottt:
It looks like the instruction is only generated with an `(unmasked)` argument for mask, which will not cause the fixed-late-use to be generated. I think that explains why we're not seeing the assertion failure for that case.

elliottt (approved):
I think this seems good, and we've got a real use in cranelift that will benefitf from this. Thanks!

What do you think about adding `reg_fixed_early_def` and `reg_fixed_late_use` functions to the `Operand` struct to simplify construction of these cases?

PR #169, opened 2023-11-24
Remove support for backwards `InstRange`

Amanieu:
These are not needed except for one place in live range calculation where we can use a reversed iterator instead.


elliottt (approved):
This is a nice simplification, thank you!

PR #171, opened 2023-12-06
Spill for VReg lifetimes instead of SpillSet lifetimes

elliottt:
This PR removes the redundant move eliminator by changing where we store spill allocations: switching from allocating spills to bundles to instead spilling vregs for their full lifetime. Changes to the cranelift test suite are for the most part positive, though there are some cases where the number of stack slots grows as a result of the new spilling strategy. I think that this could probably be mitigated by revisiting how we compute spill weight, but would also like to see what the change to more substantial targets like spidermonkey are before committing to that.

The change consists of three parts:

1. Moving the spill allocation granularity to VRegs instead of spill bundles
2. Using the new authoritative spill location to avoid writing to spill slots unnecessarily (writing when processing a def or block param instead)
3. And finally, removing the redundant move eliminator now that we don't generate unnecessary moves back into spill slots

I've fuzzed this for about 24 hours without discovering any new test failures, but I think it's a pretty substantial change that warrants more discussion once @cfallin is back.


elliottt:
I benchmarked this PR today and found that it's a pretty significant regression for the execution time of most of our benchmarks. As such, I'm going to close the PR. There could still be a path forward to removing the RME, but it's beginning to feel less worthwhile :)

PR #174, opened 2024-03-07
early break in cfg.rs

lengyijun:


bjorn3:
Can you explain why this change is made?

cfallin:
Hi @lengyijun -- thanks for the PR! It seems correct to me; the loop is scanning for a condition and sets a flag if so, with no other side-effects, so it's safe to `break` early. It's unlikely to make much of a performance difference as almost all blocks will have 1-2 successors, but there's no reason not to do it either.

I'll approve and merge, but one point to help mentor future contributions (I think this is what @bjorn3 is getting at too): it's a good idea generally to have a description for the PR as well. Even something like "I noticed this loop doesn't seem to need to continue once it sets the flag, and this could be a small optimization". Otherwise the reviewer has to dig in a bit deeper to infer the intent. Not a huge deal here, just a good habit to foster for contributions.

Thanks again, though; we're happy to see interest and more contributions on the project.

cfallin (approved):

PR #175, opened 2024-04-30
Revert "Make parallel-moves resolver available publicly"

jameysharp:
Reverts bytecodealliance/regalloc2#156

We never used this functionality in Cranelift. When we tried, we discovered that's it's pretty tricky to use correctly outside of the specific context it was written for in RA2, and then we found out that we didn't need it at all. So let's revert making it public.


cfallin (approved):

cfallin:
It looks like `cargo-deny` is failing by panic'ing (!) about a not-yet-fetched repository; I'm not sure what changed, but comparing to wasmtime's [deny.toml](https://github.com/bytecodealliance/wasmtime/blob/main/deny.toml) it looks like we don't deny `yank` there but we do here and that looks related to the error. Mind trying to edit the config to see if that fixes it?

PR #177, opened 2024-05-14
PregSet machine env

KGrewal1:
As mentioned in https://github.com/bytecodealliance/wasmtime/pull/8489#issuecomment-2106379885 , use PRegSet instead of Vecs for storing the physical registers. I'm slightly unsure if I have done this correctly for the `reg_traversal` iterator however and am unsure how it's performance will change now its calling n iterator steps as opposed to indexing straight into an array

### DIFF
--- a/src/ion/reg_traversal.rs
+++ b/src/ion/reg_traversal.rs
@@ -52,13 +52,13 @@ impl<'a> RegTraversalIter<'a> {
         }
         let hints = [hint_reg, hint2_reg];
         let class = class as u8 as usize;
-        let offset_pref = if env.preferred_regs_by_class[class].len() > 0 {
-            offset % env.preferred_regs_by_class[class].len()
+        let offset_pref = if env.preferred_regs_by_class.len_class(class) > 0 {
+            offset % env.preferred_regs_by_class.len_class(class)
         } else {
             0
         };
-        let offset_non_pref = if env.non_preferred_regs_by_class[class].len() > 0 {
-            offset % env.non_preferred_regs_by_class[class].len()
+        let offset_non_pref = if env.non_preferred_regs_by_class.len_class(class) > 0 {
+            offset % env.non_preferred_regs_by_class.len_class(class)
         } else {
             0
         };
@@ -100,9 +100,13 @@ impl<'a> core::iter::Iterator for RegTraversalIter<'a> {
             return h;
         }
 
-        let n_pref_regs = self.env.preferred_regs_by_class[self.class].len();
+        let n_pref_regs = self.env.preferred_regs_by_class.len_class(self.class);
         while self.pref_idx < n_pref_regs {
-            let mut arr = self.env.preferred_regs_by_class[self.class].into_iter();
+            let mut arr = self
+                .env
+                .preferred_regs_by_class
+                .to_preg_class(self.class)
+                .into_iter();
             let r = arr.nth(wrap(self.pref_idx + self.offset_pref, n_pref_regs));
             self.pref_idx += 1;
             if r == self.hints[0] || r == self.hints[1] {
@@ -111,9 +115,13 @@ impl<'a> core::iter::Iterator for RegTraversalIter<'a> {
             return r;
         }
 
-        let n_non_pref_regs = self.env.non_preferred_regs_by_class[self.class].len();
-        while self.non_pref_idx < self.env.non_preferred_regs_by_class[self.class].len() {
-            let mut arr = self.env.non_preferred_regs_by_class[self.class].into_iter();
+        let n_non_pref_regs = self.env.non_preferred_regs_by_class.len_class(self.class);
+        while self.non_pref_idx < n_non_pref_regs {
+            let mut arr = self
+                .env
+                .non_preferred_regs_by_class
+                .to_preg_class(self.class)
+                .into_iter();
             let r = arr.nth(wrap(
                 self.non_pref_idx + self.offset_non_pref,
                 n_non_pref_regs,
--- a/src/ion/reg_traversal.rs
+++ b/src/ion/reg_traversal.rs
@@ -1,4 +1,4 @@
-use crate::{MachineEnv, PReg, PRegClass, RegClass};
+use crate::{find_nth, MachineEnv, PReg, RegClass};
 /// This iterator represents a traversal through all allocatable
 /// registers of a given class, in a certain order designed to
 /// minimize allocation contention.
@@ -14,16 +14,14 @@ use crate::{MachineEnv, PReg, PRegClass, RegClass};
 ///   respectively, to minimize clobber-saves; but they need not.)
 
 pub struct RegTraversalIter {
-    pref_regs_by_class: PRegClass,
-    non_pref_regs_by_class: PRegClass,
-    hints: [Option<PReg>; 2],
-    hint_idx: usize,
-    pref_idx: usize,
-    non_pref_idx: usize,
-    offset_pref: usize,
-    offset_non_pref: usize,
+    pref_regs_first: u64,
+    pref_regs_second: u64,
+    non_pref_regs_first: u64,
+    non_pref_regs_second: u64,
+    hint_regs: u64,
     is_fixed: bool,
     fixed: Option<PReg>,
+    class_mask: u8,
 }
 
 impl RegTraversalIter {
@@ -35,48 +33,69 @@ impl RegTraversalIter {
         offset: usize,
         fixed: Option<PReg>,
     ) -> Self {
-        let mut hint_reg = if hint_reg != PReg::invalid() {
-            Some(hint_reg)
-        } else {
-            None
-        };
-        let mut hint2_reg = if hint2_reg != PReg::invalid() {
-            Some(hint2_reg)
-        } else {
-            None
-        };
+        // get a mask for the hint registers
+        let mut hint_mask = 0u64;
 
-        if hint_reg.is_none() {
-            hint_reg = hint2_reg;
-            hint2_reg = None;
+        if hint_reg != PReg::invalid() {
+            let mask = 1u64 << (hint_reg.bits & 0b0011_1111);
+            hint_mask |= mask;
         }
-        let hints = [hint_reg, hint2_reg];
+
+        if hint2_reg != PReg::invalid() {
+            let mask = 1u64 << (hint2_reg.bits & 0b0011_1111);
+            hint_mask |= mask;
+        }
+
         let class = class as u8 as usize;
 
-        let pref_regs_by_class = env.preferred_regs_by_class.to_preg_class(class);
-        let non_pref_regs_by_class = env.non_preferred_regs_by_class.to_preg_class(class);
+        let pref_regs_by_class = env.preferred_regs_by_class.bits[class];
+        let non_pref_regs_by_class = env.non_preferred_regs_by_class.bits[class];
 
-        let offset_pref = if pref_regs_by_class.len() > 0 {
-            offset % pref_regs_by_class.len()
+        let n_pref_regs = pref_regs_by_class.count_ones() as usize;
+        let n_non_pref_regs = non_pref_regs_by_class.count_ones() as usize;
+
+        let offset_pref = if n_pref_regs > 0 {
+            offset % n_pref_regs
         } else {
             0
         };
-        let offset_non_pref = if non_pref_regs_by_class.len() > 0 {
-            offset % non_pref_regs_by_class.len()
+        let offset_non_pref = if n_non_pref_regs > 0 {
+            offset % n_non_pref_regs
         } else {
             0
         };
+
+        // we want to split the pref registers bit vectors into two sets
+        // with the offset lowest bits in one and the rest in the other
+        let split_num = (n_pref_regs - offset_pref) as u64;
+        let split_pos = find_nth(pref_regs_by_class, split_num);
+        let mask = (1 << split_pos) - 1;
+        let pref_regs_first = pref_regs_by_class & !mask;
+        let pref_regs_second = pref_regs_by_class & mask;
+
+        let split_num = (n_non_pref_regs - offset_non_pref) as u64;
+        let split_pos = find_nth(non_pref_regs_by_class, split_num);
+        let mask = (1 << split_pos) - 1;
+        let non_pref_regs_first = non_pref_regs_by_class & !mask;
+        let non_pref_regs_second = non_pref_regs_by_class & mask;
+
+        // remove the hint registers from the bit vectors
+        let pref_regs_first = pref_regs_first & !hint_mask;
+        let pref_regs_second = pref_regs_second & !hint_mask;
+        let non_pref_regs_first = non_pref_regs_first & !hint_mask;
+        let non_pref_regs_second = non_pref_regs_second & !hint_mask;
+
+        let class_mask = (class as u8) << 6;
+
         Self {
-            pref_regs_by_class,
-            non_pref_regs_by_class,
-            hints,
-            hint_idx: 0,
-            pref_idx: 0,
-            non_pref_idx: 0,
-            offset_pref,
-            offset_non_pref,
+            pref_regs_first,
+            pref_regs_second,
+            non_pref_regs_first,
+            non_pref_regs_second,
+            hint_regs: hint_mask,
             is_fixed: fixed.is_some(),
             fixed,
+            class_mask,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


jameysharp (commented):
Awesome, thank you for tackling this!

The big picture thing to note here is that a `PRegSet` already covers all register classes, so there's no need to have an array of three of them. That should simplify things in a variety of places in this PR.

In the places where we only want to examine the registers that are within a particular class, it would be handy to add a `PRegSet::in_class` method (I'm open to suggestions for alternative names) which returns a copy where only the bits from that class are set. I can think of several ways to write that but I don't know which will be easiest.

We should put some thought into whether we can somehow restructure the loop where you're calling `nth`, to avoid needing to do that. But if we can't, there are bit-twiddling tricks for quickly finding the nth set bit. See the section immediately above https://graphics.stanford.edu/~seander/bithacks.html#ParityNaive titled "Select the bit position (from the most-significant bit) with the given count (rank)". (I can't link to that section directly for some reason.) We could override the default implementation of `Iterator::nth` with something along those lines.

There are several other iterator operations we can provide faster implementations for here. Your implementation of `n_regs` would be useful for both `Iterator::count` and `Iterator::size_hint`, and with an appropriate implementation of `size_hint` we could also implement `ExactSizeIterator` and be able to call `len()` on these iterators. I would mildly prefer to use those iterator methods instead of introducing an `n_regs` method, if you don't mind doing that in this PR. We could also provide an efficient implementation for `last` if we want to.

A change which I've been meaning to make that would make some of this PR easier is that `PRegSet` could itself implement `Iterator`, instead of having a separate `PRegSetIter` type which is basically just a copy of `PRegSet`. Then you can avoid calling `.into_iter()` in a bunch of places. Since `PRegSet` is `Copy`, it's fine for its iterator to destructively mutate it; in places where the original value is still needed it's easy to just copy it.

I hope this helps with a broad strokes plan for this PR; I may have more detailed feedback once these changes are out of the way. Most of the above is not strictly necessary for this PR, I just think it would help, so feel free to push back if you disagree!

lpereira:
To find the nth-set bit you can use things like `trailing_zeros()` and `leading_zeros()`, which will compile down to a single instruction in most targets. So, for instance, if you want to iterate over all the set bits in an 64-bit integer, you'd do something along the lines of:

```rust
let mut bitmap: u64 = ...;

while let zeros = bitmap.trailing_zeros() {
    // zeros is the index of the least-significant-bit set

    bitmap &= bitmap - 1; // or something like: bitmap &= !(1<<zeros);
}
```

jameysharp:
That's right, @lpereira, and that's the kind of code we have now: https://github.com/bytecodealliance/regalloc2/blob/b13cda1498cb001ff0cbb831e8b1d37137ad18e4/src/lib.rs#L259-L271

However, that takes time linear in the number of set bits, and there are faster alternatives if you want the index of the nth set bit.

I'm still hoping to instead find that we can rewrite the loop where this PR is calling `Iterator::nth` to not need to do that at all, but I haven't thought it through carefully enough to see how to do that yet. Perhaps you'd like to take a look at the changes in this PR to src/ion/reg_traversal.rs and see if you have suggestions?

KGrewal1:
Thanks for the feedback!

As the `count_ones` function does not modify the  original, wouldn't it be suitable for `len` in `ExactSizeIterator`. 

Regarding an `in_class` method, and having only `PRegSet`, as opposed to `[PRegSet, 3]`, I'm unsure where the class data would then be stored in `MachineEnv`?

Have updated to implement `Iterator` and `ExactSizeIterator` directly on `PRegSet`. Still need to do more efficient implementations of `nth` and `last`

cfallin:
I think it might be good to take a step back here: why are we considering bitsets to compact the set? What is the goal?

From the linked comment above, we'd like to have a statically-initialized `MachineEnv` if we can, and dynamically-allocated lists of registers preclude that. But as noted here, having the list of registers to iterate over during the inner allocation loop is fairly performance-critical. Ordinarily one would want to consider a bitset when storage density is more important than simple/fast iteration.

But I think there's a better way: why not take a borrow to a `'static` slice for the register list instead of a `PRegSet`? That allows us to have a `MachineEnv` that is fully static, but at the same time preserves the existing algorithms (and the simplicity of implementation).

Alternately, there is [this suggestion](https://github.com/bytecodealliance/wasmtime/pull/8489#issuecomment-2091875500) I left on the original thread. I'm not sure if that got overlooked or lost -- but allocating the `MachineEnv` once when the backend is set up (normally once in the process for `wasmtime compile` or similar) is another design point that preserves most of the existing benefits but avoids the need to make `MachineEnv` statically-initializable (which might be a hurdle if we add more complex state in the future).

Either of these could work, and both I think are preferable to complex Hacker's Delight bit twiddling tricks and compromises in iteration speed...

KGrewal1:
Thanks: yes the confusion was regarding how PRegSet encoded the types of register, this clears it up

cfallin (changes requested):
(r-'ing this PR until above points about why `PRegSet` at all are addressed; I don't think we should use it at all)

jameysharp:
There are two big-picture questions to examine here:
1. How do we make accessing an appropriate `MachineEnv` fast when we need it? As @cfallin pointed out, this is the question which is relevant to bytecodealliance/wasmtime#8489.
2. What representation of `MachineEnv` should we use to support the use cases we have for it inside regalloc2?

Regarding the first question, I agree with Chris: if we can construct the `MachineEnv` once up front, maybe in frontends like Wasmtime or rustc-codegen-clif, that would be ideal. I'm not sure that is feasible if the set of allocatable registers depends on the calling convention of individual functions, though. In fact I don't understand why the current backends _don't_ depend on calling convention: why is it okay to always use SysV's preferred and non-preferred register sets on x64, even on Windows?

If that doesn't work, it's certainly also possible to require that the registers be in a slice with `'static` lifetime. If that's too restrictive for some users of regalloc2, a `Cow<'static, [PReg]>` would allow choosing dynamically between using `Vec<PReg>` or `&'static [PReg]`.

I think there is still an interesting conversation to have about the second question. I believe `RegTraversalIter` is the key point of contention here. (The conversion from `MachineEnv` to `PRegSet` is obviously easier if we use `PRegSet`, and was what originally motivated me to consider this, but it's also not performance critical since bytecodealliance/wasmtime#8457 landed.) Since I'm trying to learn how this is used today, I'm going to write down how I think this works first.

When `RegTraversalIter::new` is called, it constructs an iterator that returns registers in this order:
- if `fixed` was provided, then that is the only register that will be returned; otherwise:
- `hint_reg`, if provided
- `hint2_reg`, if provided
- all the preferred registers in the class, excluding the hint registers, but the list is rotated by `offset`
- all the non-preferred registers in the class, excluding the hint registers, but the list is rotated by `offset`

Rotating a list by an offset is equivalent to splitting the list at that offset, and iterating over the second part before the first one. We certainly _could_ duplicate that behavior efficiently with `PRegSet`: splitting a bitset into two parts at a given offset is easy enough, and then we can pull the next element from each part in constant time using only the iterator we already have. We can also remove the hint registers from the sets up front rather than having to check for them on each iteration.

I _think_ this is both simpler and faster than the current implementation. With some optimization, `RegTraversalIter` shrinks from 64 bytes to something like 40 or 48 bytes, I think we need just a few instructions per call to `next()`, and LLVM could plausibly keep the whole state in registers. The specifics depend on questions like: "does `hint_reg` always need to be returned before `hint2_reg`, or is it okay if they get swapped?" and "how exactly do we want the set of registers to be split according to `offset`?" but in broad strokes, I think this could be worth doing.

cfallin:
Thanks for writing all this down, Jamey -- I agree with your assessment that actually a bitset could work here with an O(1) step, with the split-the-bitset-in-half-at-the-rotation-point approach.

I think my main point of confusion through all of this discussion -- prior to the above -- has been trying to understand the motivation; my initial sense was that this would be a loss both in performance (complex bit-twiddling for linear-time access of Nth element) and an increase in complexity, to pay for... constness that can be had other ways too? But given the above at least the linear-time concern is gone.

The question I'd ask of any new approach is twofold: does it increase clarity of implementation, and/or does it make a measurable reduction in runtime? (Either alone is IMHO good justification for a change. Conversely, an implementation that's about the same complexity and has no change in performance isn't worth the effort and risk to undergo.) My guess for the latter (based only on experience profiling this loop 2-3 years ago) is that the traversal iter implementation doesn't matter all that much; there's a ton of logic in the loop (the btree lookup and liverange collision checks) that takes the majority of the time. But it's possible that I'm wrong about that. And we'd want to see an implementation (with the new approach) to evaluate the former.

KGrewal1:
If `PRegSet` is used for `RegTraversalIter`, where the out of order access occurs (and so is likely to be slower and/or have some bithackery), what is the advantage of the keeping another representation in `MachineEnv`, as the use in ion/process in which it is iterated in order over, and liveranges in which only `last` is called should be quite efficient and relatively simple code (especially as these are only done a class at a time and so could be further sped up by using the fact that int registers are first 64 bits, floats second and vector third, which wasn't used in this PR). 

Definitely think this has significantly broader implications than I'd intended however.

jameysharp:
I've had this on my to-do list to investigate since working on bytecodealliance/wasmtime#8457, because at a quick glance it didn't make sense to me to use arrays of `PReg`s… pretty much anywhere. A `PRegSet` is basically the same size as an empty `Vec`, and does no heap allocation. But I hadn't looked into the details.

Now that I better understand how `MachineEnv` is used in regalloc2, I'm still interested in doing this experiment and will leave it on my to-do list if nobody else tries it first, but there's definitely a bigger obligation than I expected to show a benefit from the change.

> Definitely think this has significantly broader implications than I'd intended however.

Yeah, me too! If you would prefer to try out Chris' suggestions (or my variations) about other ways to remove the `OnceLock` in Cranelift, that would be great too.

> If `PRegSet` is used for `RegTraversalIter`, where the out of order access occurs (and so is likely to be slower in this PR), what is the advantage of the keeping another representation in `MachineEnv`, as the use in ion/process in which it is iterated in order over, and liveranges in which only `last` is called should be quite efficient and relatively simple code (especially as these are only done a class at a time and so could be further sped up by using the fact that int registers are first 64 bits, floats second and vector third, which wasn't used in this PR).

I don't understand this question. I think we can make it fast and simple to use `PRegSet` (or a variation of the same idea) in `RegTraversalIter`, and if that's true, then I think it's reasonable to also use `PRegSet` in `MachineEnv`. And I think Chris said above that if making this change makes the code either significantly faster or easier to understand, then it's worth doing. Does that answer your question?

KGrewal1:
if that's true, then I think it's reasonable to also use `PRegSet` in `MachineEnv`. 

Thanks, yes this does answer my question

KGrewal1:
@jameysharp I think this is close to ready, still a bit unsure on the select n-th code from https://graphics.stanford.edu/~seander/bithacks.html , but I think it has somewhat simplified `reg_traversal_iter` code.

src/lib.rs
```diff
@@ -381,6 +381,64 @@ impl Iterator for PRegClass {
...
+// find the n-th set bit from the RIGHT,
+// using 1 based indexing
+// returns distance from the LEFT
+fn find_nth(v: u64, mut r: u64) -> u64 {
+    const C: u64 = 0b00000000_11111111_00000000_11111111_00000000_11111111_00000000_11111111; // 0x00FF00FF
+    const D: u64 = 0b00001111_00001111_00001111_00001111_00001111_00001111_00001111_00001111; // 0xF0F0F0F0
+    const E: u64 = 0b00110011_00110011_00110011_00110011_00110011_00110011_00110011_00110011; // 0x33333333
+    const F: u64 = 0b01010101_01010101_01010101_01010101_01010101_01010101_01010101_01010101; // 0x55555555
+
+    // from https://graphics.stanford.edu/~seander/bithacks.html: uses 1 based indexing
```
bjorn3, line 413: Which of the algorithms on that page is this?

KGrewal1: Select the bit position (from the most-significant bit) with the given count (rank) 
Unfortunately the links to the individual algorithms don't seem to work

bjorn3: https://graphics.stanford.edu/~seander/bithacks.html##SelectPosFromMSBRank works. This page (likely accidentally) used `#SelectPosFromMSBRank` as name rather than `SelectPosFromMSBRank` so you need two hashes.

KGrewal1: Thanks, changed the link

jameysharp (commented):
I like how this is coming together! Thanks for your patience waiting for me to give this a detailed review.

Personally, I think you've demonstrated that we can address @cfallin's concerns, but I won't speak for him. Chris, could you check whether you're still concerned about overall algorithmic complexity with this version?

I have a number of suggestions for simplifying this PR further which I think are important for keeping this code as maintainable as we can, but overall I'm in favor of these changes. Thank you for your efforts!

cfallin:
I think this is going in the right direction for sure -- thanks @KGrewal1 for your continued work on this despite my earlier skepticism!

Seeing the full glory of `find_nth`, I do have to admit I find myself fairly concerned. It's very complex code (even if transcribed from a cited algorithm online), if we ship it we should definitely include in-depth tests for it, and it's quite a bit heavier-weight than I had imagined previously -- I count 79 ALU ops, and we do that twice per traversal setup, which happens once per liverange (potentially many times per original liverange). Allocator efficiency should be something like a few thousand instructions per liverange, so that's really nontrivial: I think it is warranted to do a full benchmarking to make sure it doesn't impact the common case of short functions.

However I think we can actually avoid it altogether, along with the modulo operator in the `RegTraversalIterator` setup, using one of several approaches.

The thought goes something like: I had originally added the "offset and wraparound" behavior to `RegTraversalIter` to ensure that various allocations try probing registers in an evenly-distributed way, rather than biasing toward the early elements in the list, to minimize contention in the allocation maps. The effect on allocation speed was pretty dramatic: if memory serves, something like an average of 13 BTree probes down to an average of 4 probes per LiveRange processing step. (This at the cost of "spreading out a bit and enjoying the room" in the register file, rather than neatly packing r0, r1, r2, ... tightly.)

But there's no reason we need to preserve the *exact* traversal order and distribution scheme; all that's needed is some distribution, and it need not even be perfectly spread.

- The simplest one I can think of is: take `offset mod 128` (i.e. `offset & 127`); that's the splitpoint. Effectively we're rotating the masks one position per instruction-offset rather than one *set element* per instruction-offset. If PReg indices in the set are fairly evenly distributed, this will have the same "load-spreading" effect.
- However if indices are clustered (say, reg indices 0..24 are in the set, and 25..128 are not), then we have a bias in the probing distribution (in this example, 128-24 = 104 out of every 128 instructions will start at p0).
- But we can push back against that expected clustering with a "segmented" splitting strategy: for example, take N = `offset mod 32` or `offset mod 16` and, in every 32- or 16-bit segment of the set, split at bit N. I think this would work something like (using u128s for clarity rather than two u64s):

```
let split = offset & 31;
let mask = (0x00000001_00000001_00000001_00000001_u128 << split) - 1;
let bits1 = bits & mask;
let bits2 = bits & !mask;
```

  This still has a bias but not as bad; it works out well because the segment size of 32 elements is just a bit bigger than the 0..24 register range. In practice I suspect a segment size of 32 bits may work well enough.

  We can evaluate this by checking the `process_bundle_reg_probes_any` stat in mainline vs. this change.

- Generalizing a bit, we could do two or more such steps: split at some bit position within 32-bit segments, then again within 16-bit segments (then with ...). This is kind of like a bitonic sort network but running backward, to give some arbitrary subset. I can write this out if interested but I'd be curious how well the above does first.

- Another thought that comes to mind: there's no reason we can't keep some state in the env, such that each time we do a traversal, we start at the next position. This means that we're doing the offsetting by "processing step number" rather than instruction offset but no matter. In effect we can get the starting position for traversal N+1 by taking the starting position for traversal N, taking the lowest-set-bit and removing it from first-half and putting it in second-half, and flipping them when the first-half becomes 0.

Anyway, let me know what you think! If I had to choose I think I'd be interested in seeing the last one implemented as it keeps the even distribution of probes perfectly while doing only a few clz/ctz ops per traversal step and per iter step.

src/ion/reg_traversal.rs
```diff
@@ -14,109 +13,142 @@ use crate::{MachineEnv, PReg, RegClass};
...
+        let split_pos = find_nth(non_pref_regs_by_class, split_num);
+        let mask = (1 << split_pos) - 1;
+        let non_pref_regs_first = non_pref_regs_by_class & !mask;
+        let non_pref_regs_second = non_pref_regs_by_class & mask;
+
+        // remove the hint registers from the bit vectors
+        let pref_regs_first = pref_regs_first & !hint_mask;
+        let pref_regs_second = pref_regs_second & !hint_mask;
+        let non_pref_regs_first = non_pref_regs_first & !hint_mask;
+        let non_pref_regs_second = non_pref_regs_second & !hint_mask;
```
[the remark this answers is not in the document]
KGrewal1: Yes, I'd done it this way in order to be as close to the original split as possible: I may try they way Chris suggested, that doesn't depend on doing a bit trick to find the split point instead as it does seem 'simplest'

src/lib.rs
```diff
@@ -191,106 +197,253 @@ impl core::fmt::Display for PReg {
...
+    fn from(regs: Vec<PReg>) -> Self {
+        let mut res = Self::default();
+
+        for preg in regs {
+            res.add(preg);
         }
 
         res
     }
 }
```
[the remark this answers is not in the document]
KGrewal1: Yes, I'll remove it

src/ion/reg_traversal.rs
```diff
@@ -14,109 +13,142 @@ use crate::{MachineEnv, PReg, RegClass};
...
-                idx - limit
-            } else {
-                idx
-            }
+        // if there are hints, return them first
+        if self.hint_regs != 0 {
+            let index = self.hint_regs.trailing_zeros() as u8;
+            self.hint_regs &= !(1u64 << index);
+            let reg_index = index as u8 | self.class_mask;
+            return Some(PReg::from(reg_index));
```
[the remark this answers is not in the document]
KGrewal1: Hmmm I'd actually done this for the opposite reason: I found it easier to reason/think about what was occurring when I could think of it as a 2 bit number for the class and 6 bit number for the position in the class, though that also requires fully committing to that representation, as opposed to mentally going through the associated constants and casts between types (I'm not actually sure how much leeway there is to change the bitpacked representation as doing it this way definitely would concrete it as it is)

KGrewal1:
Thanks for the feedback: just finished my finals, so will try do a PR to address them this weekend

jameysharp:
Thanks for writing all that up, Chris! I found it very helpful. I think in various places where you said "128" you probably meant "64", right? That's how many PRegs we have in each class right now. The `mask` expression also probably doesn't do what you intended but we can figure that out.

> Thanks for the feedback: just finished my finals, so will try do a PR to address them this weekend

Congratulations, and I'd suggest taking some time to relax! I'll be pretty busy next week so I may not be able to respond that quickly anyway.

cfallin:
>  The mask expression also probably doesn't do what you intended but we can figure that out.

Err, yes, sorry, I think I meant `... - 0x00000001_00000001_00000001_00000001` rather than `- 1` (basically I want a mask like `0b0000..1111..0000..1111..0000..1111..0000..1111` where the groups are the bottom halves of 32-bit lanes)

I do very much think the last idea is the way to go though (keeping state, starting where we left off and rotating one place) both for simplicity and performance, so hopefully the above isn't actually needed :-)

KGrewal1:
Really sorry but started an internship and don't think I'm going to be able to finish this in the next few months😢. All this work was done whilst I was a student, so it's available if anyone else wants to finish it/use it to build off.

KGrewal1:
This seems to be unnecessary given greater refactors in PRegSet now anyway

PR #178, opened 2024-06-01
Refactor PRegSet.

numas13:
The changes will make it easier to increase the number of physical registers and register classes in the future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -183,6 +183,9 @@ impl core::fmt::Display for PReg {
     }
 }
 
+/// A type for internal bit arrays.
+type Bits = u64;
+
 /// A physical register set. Used to represent clobbers
 /// efficiently.
 ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -191,82 +194,92 @@ impl core::fmt::Display for PReg {
...
-        Self { bits: [0; 2] }
+        Self {
+            bits: [0; Self::LEN],
+        }
+    }
+
+    /// Splits the given register index into parts to access the internal bit array.
+    const fn split_index(reg: PReg) -> (usize, usize) {
+        let index = reg.index();
+        (index / Self::BITS, index % Self::BITS)
```
cfallin, line 217: It should be optimized via strength reduction but just to be sure in unoptimized builds: could we use shifts and masking here instead? AFAIK, `BITS` will always be a power of two.

src/lib.rs
```diff
@@ -191,82 +194,92 @@ impl core::fmt::Display for PReg {
...
-            let index = self.bits[1].trailing_zeros();
-            self.bits[1] &= !(1u128 << index);
-            Some(PReg::from_index(index as usize + 128))
-        } else {
-            None
+        loop {
+            let bits = self.bits.get_mut(self.cur)?;
+            if *bits != 0 {
+                let bit = bits.trailing_zeros();
+                *bits ^= 1 << bit;
```
cfallin, line 278: Can we stick with the `&= !(1 << bit)` form we had before, here? Equivalent in this case and not a huge deal either way, just a preference, but this form for me is slightly more clear in intent (clearing a bit, vs. XOR which I ordinarily see for flipping the bit).

cfallin (approved):
Thanks very much for this contribution! Two minor-nit comments below reflecting some preferences but nothing major; happy to have this generalization.

PR #179, opened 2024-07-19
Added support for multiple OperandConstraint::Reuse operands.

Iizerd:
Previously instructions that have multiple operands specifying the constraint `OperandConstraint::Reuse` would not work properly. Good examples of this are X86's `XCHG` and `XADD`, both of which require two `OperandConstraint::Reuse`s. This PR addresses this issue and fixes it by using a `SmallVec<[VReg; 4]>` to store all reuse constraints to check, instead of a single `Option<VReg>`.

I ran 8 instances of ion_checker overnight without fail. Additionally, the fixes have enabled my compiler to correctly emit the aforementioned instructions and it passes my own test suite. I'm not however confident that the ion_checker correctly generates these types of instructions as prior to this change it was still reporting no errors. It also seems to only generate one Def per instruction.

I will be leaving a comment in #145 as I discovered some interesting things about that as well while looking into this.

### DIFF
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -460,15 +460,14 @@ impl<'a, F: Function> Env<'a, F> {
                 // Does the instruction have any input-reusing
                 // outputs? This is important below to establish
                 // proper interference wrt other inputs. We note the
-                // *vreg* that is reused, not the index.
-                let mut reused_input = None;
+                // *vreg*s that are reused, not the index.
+                let mut reused_inputs: SmallVec<[VReg; 4]> = smallvec![];
                 for op in self.func.inst_operands(inst) {
                     if let OperandConstraint::Reuse(i) = op.constraint() {
                         debug_assert!(self.func.inst_operands(inst)[i]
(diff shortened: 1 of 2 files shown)


cfallin:
Thanks for looking into this! The change does look correct to me.

Regarding fuzzing -- I believe you're correct that we actually only add one def per instruction and up to one reuse of that def when generating test functions (see [here](https://github.com/bytecodealliance/regalloc2/blob/f4812a23efc82466ed8377c6e11113134e13f425/src/fuzzing/func.rs#L446)). It would be good to see if we can relax that restriction (on both counts) and test with that.

cfallin:
@lizerd any particular reason you've closed this PR? We're happy to take the change, with the fuzzing changes as suggested above!

Iizerd:
I realized I should have created a separate branch with just the one commit that I made relating to that. I did not realize that future commits(as seen above) would also get added to the pr 👀 whoops! I can reopen it but let me create a separate branch that only has the one commit first that actually makes the change mentioned.

PR #181, opened 2024-08-05
Fastalloc1

d-sonuga:
This is the initial implementation of the fast register allocator in `src/fastalloc`.
It's still a work in progress and I haven't done any kind of optimizations on it. Still using less-than-optimal data structures and stuff.

When I was initially developing it, I used solely the fuzzer to give me the cases to think about and to determine correctness, and I thought I was done with getting the algorithm correct after running the fuzzer for a few days without any errors. But after trying to measure its performance with sightglass and watching it crash several times, I've rethought through the code and reworked it and I've run it through the fuzzer for a few hours now, but it still crashes on some of the benchmarks, and that's what I'm working on resolving now.

I'm making this draft PR now to get some feedback on what I've done so far and on the algorithm itself.


src/lib.rs
```diff
@@ -1558,4 +1563,7 @@ pub struct RegallocOptions {
 
     /// Run the SSA validator before allocating registers.
     pub validate_ssa: bool,
+
+    /// Run the SSRA algorithm
+    pub use_fastalloc: bool,
```
cfallin, line 1568: Rather than a `bool`, can we define an `enum Algorithm { Ion, Fastalloc }` and take an option `pub algorithm: Algorithm`? This leaves room for adding others in the future and makes the intent more clear.

(We can keep the `Default` derivation working by adding `#[default]` to the `Ion` arm of the enum)

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1793 @@
...
+        }
+        trace!("");
+    }
+
+    let mut env = Env::new(func, mach_env);
+    env.run()?;
+
+trace!("Final edits: {:?}", env.edits);
+trace!("safepoint_slots: {:?}", env.safepoint_slots);
+trace!("\n\n\n\n\n\n\n");
```
cfallin, line 1782: minor nit, but `cargo fmt` should indent these statements properly

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1793 @@
...
+            self.alloc_block(Block::new(block));
+        }
+        self.build_safepoint_stackmap();
+
+        /////////////////////////////////////////////////////////////////////////////////////
+        trace!("Done!");
+        struct Z(usize);
+        impl std::fmt::Debug for Z {
+            fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                write!(f, "v{}", self.0)
```
cfallin, line 1722: I think the `VReg` should have an equivalent `Display` implementation, so instead we could push `VReg::new(i, RegClass::Int)` below (class isn't printed by `Display`)

src/fastalloc/iter.rs
```diff
@@ -0,0 +1,643 @@
...
+#[derive(Clone, Copy, PartialEq)]
+enum OperandConstraintKind {
+    Any,
+    Reg,
+    Stack,
+    FixedReg,
+    Reuse,
+}
+
+impl PartialEq<OperandConstraint> for OperandConstraintKind {
```
cfallin, line 12: Rather than this impl, can we implement `From<OperandConstraint> for OperandConstraintKind`, then one could do `constraint.kind() == OperandConstraintKind::Reg`? This is slightly more general and cleaner in my experience...

src/fastalloc/iter.rs
```diff
@@ -0,0 +1,643 @@
...
+            OperandConstraint::Reg => *self == Self::Reg,
+            OperandConstraint::Stack => *self == Self::Stack,
+            OperandConstraint::FixedReg(_) => *self == Self::FixedReg,
+            OperandConstraint::Reuse(_) => *self == Self::Reuse,
+        }
+    }
+}
+
+#[derive(Clone, Copy, PartialEq)]
+struct SearchConstraint {
```
cfallin, line 25: This is a neat abstraction and is certainly providing some benefit by generalizing all the cases below. However I worry a little bit about efficiency here: we're explicitly building a data structure in memory to define what kind of operands we're matching, and then dispatching based on this (iterating, in some cases). Data-driven control flow that is "constant in practice" can sometimes be optimized but often not. It may be simpler and faster to do something like:

```
impl<'a> Operands<'a> {
  fn matching<F: Fn(Operand)>(&self, predicate: F) -> impl Iterator<Item = Operand> + 'a {
    self.iter().filter(|op| predicate(*op))
  }
}
```

then

```
impl<'a> Operands<'a> {
  fn non_fixed_non_late(&self) -> impl Iterator<Item = Operand> + 'a {
    self.matching(|op| op.pos() == ... && op.constraint().kind() != ...))
  }
}
```

This offers a little more flexibility (vs. e.g. the need to have two slots for "must not have") and should inline nicely into a filtered for-loop.

src/fastalloc/lru.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Index of the most recently used register.
+    pub head: usize,
+    /// Class of registers in the cache.
+    pub regclass: RegClass,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct LruNode {
+    /// The previous physical register in the list.
+    pub prev: usize,
```
cfallin, line 26: These can almost certainly be `u16` and we'll save a little memory (so cache locality) by doing so...

src/fastalloc/lru.rs
```diff
@@ -0,0 +1,240 @@
...
+        if i == prev_newest {
+            return;
+        }
+        if self.data[prev_newest].prev != i {
+            self.remove(i);
+            self.insert_before(i, self.head);
+        }
+        self.head = i;
+        trace!("Poked {:?} in {:?} LRU", preg, self.regclass);
+        self.check_for_cycle();
```
cfallin, line 67: This could probably be annotated with `#[cfg(debug_assertions)]` -- that will enable this call when we have assertions compiled in, e.g. for fuzzing or in a general debug build, but avoid it in release builds where performance matters. I suspect that this is probably a reasonably significant slowdown right now if it's running in all your performance tests!

(likewise below too)

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1793 @@
...
+struct Allocs {
+    allocs: Vec<Allocation>,
+    /// `inst_alloc_offsets[i]` is the offset into `allocs` for the allocations of
+    /// instruction `i`'s operands.
+    inst_alloc_offsets: Vec<u32>,
+}
+
+impl Allocs {
+    fn new<F: Function>(func: &F, env: &MachineEnv) -> Self {
+        // The number of operands is <= number of virtual registers
```
cfallin, line 30: I don't think this is actually true -- we could have more operands than virtual registers because of uses (imagine we have a function with a single vreg `v1`, and a instruction that uses `v1` a million times). It shouldn't really matter here since this is just a pre-allocation heuristic; but it may be better to do something like `func.num_insts() * 3` as a first guess.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1793 @@
...
+
+    /// The current allocations for all virtual registers.
+    vreg_allocs: Vec<Allocation>,
+    /// Spillslots for all virtual registers.
+    /// `vreg_spillslots[i]` is the spillslot for virtual register `i`.
+    vreg_spillslots: Vec<SpillSlot>,
+    /// The virtual registers that are currently live.
+    live_vregs: HashSet<VReg>,
+    /// Allocatable free physical registers for classes Int, Float, and Vector, respectively.
+    freepregs: PartedByRegClass<BTreeSet<PReg>>,
```
cfallin, line 76: as Amanieu mentioned earlier, a `PRegSet` should be much more efficient here (and will allow us to replace some iteration below with intersection/union operations I think).

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1793 @@
...
+    /// allocating a use operand.
+    vregs_first_seen_in_curr_inst: HashSet<VReg>,
+    /// Used to keep track of which vregs have been allocated in the current instruction.
+    /// This is used to determine which edits to insert when allocating a use operand.
+    vregs_allocd_in_curr_inst: HashSet<VReg>,
+    /// All the safepoint instructions encountered during allocation and their blocks.
+    /// When allocation is completed, this contains all the safepoint instructions
+    /// in the function.
+    /// This is used to build the stackmap after allocation is complete.
+    safepoint_insts: Vec<(Block, Inst)>,
```
cfallin, line 139: Also as we noted in the call -- we can remove safepoint-related logic as we're likely to switch over to a non-regalloc-provided safepoint implementation in Cranelift soon. This should simplify things quite a bit!

cfallin (commented):
Some initial comments from a pass over most of the "peripheral" code -- everything but the core in fastalloc/mod.rs. I'll continue my detailed pass over that and have some more thoughts later.

Overall impression: a very good start! I'm happy to see some nice abstractions built up, and the overall shape of the code looks like what I would expect. We should be pretty close to able to land the initial version, with some notes here and whatever Amanieu sees addressed, and with the correctness issue you're chasing now fixed.

Thanks again for all the work on this!

src/fastalloc/lru.rs
d-sonuga: Is there any reason why it can't be a `u8`?

src/fastalloc/mod.rs
d-sonuga: Ah, yes. This is an outdated comment.

src/fastalloc/lru.rs
cfallin: I had been thinking about the total number of possible `PReg`s and for some reason had thought we could have more than 256 of them across all classes; but (i) we actually can't, since the entire `PReg` is packed into 8 bits, and (ii) your LRU data structure is per-class anyway. So given both those reasons it can definitely be a `u8`!

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+        trace!("Done adding edits for clobbered registers");
+    }
+
+    /// If instruction `inst` is a branch in `block`,
+    /// this function places branch arguments in the spillslots
+    /// expected by the destination blocks.
+    /// 
+    /// The process used to do this is as follows:
+    /// 
+    /// 1. Move all branch arguments into corresponding temporary spillslots.
```
Amanieu, line 1394: This shouldn't be needed: because each vreg is assigned to a separate spillslot and each block parameter definition at the start of a block must be a distinct vreg (due to SSA), you're guaranteed to never have cycles. So you can just perform all the moves directly.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+        for succ in self.func.block_succs(block).iter() {
+            self.alloc_slots_for_block_params(*succ);
+        }
+
+        for (succ_idx, _) in self.func.block_succs(block).iter().enumerate() {
+            // Move from branch args spillslots to temporaries.
+            //
+            // Consider a scenario: block X branches to block Y and block Y branches to block X.
+            // Block Y has block param vp and block X uses virtual register va as the branch arg for vp.
+            // Block X has an instruction that uses vp.
```
Amanieu, line 1424: It's impossible for Block X to use vp because it isn't dominated by the definition of vp which is in Block Y. SSA requires that there only be one def per vreg and that the def dominates all uses. A block argument counts as a def of that vreg.

src/fastalloc/lru.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.validate_lru();
+    }
+
+    pub fn append_and_poke(&mut self, preg: PReg) {
+        self.append(preg.hw_enc());
+        self.poke(preg);
+    }
+
+    /// Insert node `i` before node `j` in the list.
+    pub fn insert_before(&mut self, i: u8, j: u8) {
```
Amanieu, line 134: This should be private, it's only used in this file.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+                let temp = Allocation::stack(temp_slot);
+                self.vreg_allocs[vreg.vreg()] = temp;
+                next_temp_idx[vreg.class()] += 1;
+                trace!(" --- Placing branch arg {:?} in {:?}", vreg, temp);
+                trace!("{:?} which is now in {:?} inserting move to {:?}", vreg, temp, param_alloc);
+                self.add_move_later(inst, temp, param_alloc, vreg.class(), InstPosition::Before, false);
+            }
+        }
+
+        reset_temp_idx(&mut next_temp_idx);
```
Amanieu, line 1481: This seems unnecessary, next_temp_idx isn't used anymore in this function.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+
+        for (succ_idx, _) in self.func.block_succs(block).iter().enumerate() {
+            for vreg in self.func.branch_blockparams(block, inst, succ_idx).iter() {
+                // All branch arguments should be in their spillslots at the end of the function.
+                self.vreg_allocs[vreg.vreg()] = Allocation::stack(self.vreg_spillslots[vreg.vreg()]);
+                self.live_vregs.insert(*vreg);
+            }
+        }
```
Amanieu, line 1489: This can be merged into the previous loop.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+    /// the allocations where they are expected to be before the first
+    /// instruction. 
+    fn reload_at_begin(&mut self, block: Block) {
+        trace!("Reloading live registers at the beginning of block {:?}", block);
+        trace!("Live registers at the beginning of block {:?}: {:?}", block, self.live_vregs);
+        trace!("Block params at block {:?} beginning: {:?}", block, self.func.block_params(block));
+        // We need to check for the registers that are still live.
+        // These registers are livein and they should be stack-allocated.
+        let live_vregs = self.live_vregs.clone();
+        for vreg in live_vregs.iter().cloned() {
```
Amanieu, line 1628: Since this is done at the end of processing for a block, it's faster to just clear end reset all the state at once at the end rather than freeing each allocation individually.

This function could be much simpler:
- remove all blockparams from live_vregs
- for each live vreg move it to its spillslot.

Additionally, you could change the definition of live_vregs to exclude vregs that are in a spillslot. They don't require any special handling here.

Amanieu (commented):
Some initial feedback, I'm still looking through the code.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1839 @@
...
+        for (succ_idx, _) in self.func.block_succs(block).iter().enumerate() {
+            for vreg in self.func.branch_blockparams(block, inst, succ_idx).iter() {
+                // All branch arguments should be in their spillslots at the end of the function.
+                self.vreg_allocs[vreg.vreg()] = Allocation::stack(self.vreg_spillslots[vreg.vreg()]);
+                self.live_vregs.insert(*vreg);
+            }
+        }
+    }
+
+    fn alloc_inst(&mut self, block: Block, inst: Inst) -> Result<(), RegAllocError> {
```
Amanieu, line 1492: The order of operations here seems to be incorrect since it could fail to allocate an instruction with an early-fixed-def if that register was selected by a late-nonfixed-def that was previously processed.

You can take advantage of the fact that early defs and late defs are not allowed to overlap. You can think of it as the live range of a def always extends to the end of the instruction, no matter whether it is early or late. The same applies to late uses: the vreg comes from a previous instruction so it is "in use" in both the early and late part of the instruction.

A better ordering of operations would be to separately track which registers are available in the early and late part of the instruction.
- First process all (early & late, use & def) fixed constraints & clobbers. This removes the fixed registers from the set of available registers. For early use it only removes registers from the early set. For late def it only removes registers from the late set. For early def & late use it removes from both.
- Then process all remaining defs & uses, assigning from the remaining available registers.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1839 @@
...
+            trace!("Allocation for instruction {:?} and operand {:?}: {:?}", inst, op, self.allocs[(inst.index(), op_idx)]);
+        }
+        self.vregs_allocd_in_curr_inst.insert(op.vreg());
+        if let Some(preg) = self.allocs[(inst.index(), op_idx)].as_reg() {
+            self.pregs_allocd_in_curr_inst.add(preg);
+        }
+        Ok(())
+    }
+
+    fn process_reuse_operand_allocation(
```
Amanieu, line 882: The logic for reused operands can be simplified if you think about it this way:
- Handle a reuse-def by simply copying the constraint from the use. If it's fixed, treat it as a fixed def. Otherwise treat it as a non-fixed def.
- Handle the reused use by treating it as a fixed use that uses the registers you selected for the def.

This works because reuse-def is only allowed on a late def and the reused operand must be an early use.

src/fastalloc/lru.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+        Self {
+            head: if regs.is_empty() { u8::MAX } else { regs[0].hw_enc() as u8 },
+            data,
+            regclass,
+        }
+    }
+
+    /// Marks the physical register `i` as the most recently used
+    /// and sets `vreg` as the virtual register it contains.
```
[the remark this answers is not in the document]
d-sonuga: Outdated comment....

src/fastalloc/mod.rs
d-sonuga: It is possible to have cycles because the branch arguments need to be moved into the block parameter spillslots.
In the case where some or all of the branch arguments are a permutation of the block's parameters, like:
```
block1(v1, v2)
    1. goto block2

block2
    2. goto block1(v2, v1)
```
If the moves are inserted directly for block 2, then we'll have:

```
move from stack_v2 to stack_v1
move from stack_v1 to stack_v2
```
Which is incorrect.

src/fastalloc/mod.rs
d-sonuga: Whoops, I think I was unclear there. I meant that Y dominates X and X has a loop back to Y. Something like:

```
block entry
  goto Y(defines some value for vp)

block Y(vp)
  goto X

block X
  goto Y(va)
```
So we have: entry -> Y -> X then back to -> Y

src/fastalloc/mod.rs
d-sonuga: > The order of operations here seems to be incorrect since it could fail to allocate an instruction with an early-fixed-def if that register was selected by a late-nonfixed-def that was previously processed.

The algorithm allocates late defs first, then deallocates them.
All late defs are freed before the early fixed defs are processed, so this isn't possible.
I don't think order is incorrect.

src/fastalloc/mod.rs
d-sonuga: Yes, that is a much better way to handle it!

src/fastalloc/mod.rs
Amanieu: Remember that early defs are live until the end of the instruction (the def starts at the early point, but also cover the late point). So it's invalid to assign the same register to an early def and a late def because the live ranges overlap.

d-sonuga: Oh, I thought an early def that isn't used simply doesn't have its liverange extended anywhere, allowing its allocation to be free for a late operand.
Okay, that makes sense.

src/fastalloc/mod.rs
Amanieu: You are correct, it is actually possible to cause a cycle here.

src/fastalloc/mod.rs
d-sonuga: > Since this is done at the end of processing for a block, it's faster to just clear end reset all the state at once at the end rather than freeing each allocation individually.

I don't think it's that straightforward. `freealloc` does some other things, too, that isn't just clearing stuff.

> This function could be much simpler:

    remove all blockparams from live_vregs
    for each live vreg move it to its spillslot.

Yeah, I'll definitely do that.

> Additionally, you could change the definition of live_vregs to exclude vregs that are in a spillslot. They don't require any special handling here.

And this will require an additional check after every operand allocation to see if it's a stack allocation. So, I'm not sure if the current approach will be faster?

src/fastalloc/iter.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use alloc::vec::Vec;
+    use alloc::vec;
+    use crate::{PReg, RegClass};
+    use super::*;
+
+    // Using a new function because Operand::new isn't a const function
```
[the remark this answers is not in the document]
Amanieu: If you change `debug_assert_eq!` to `debug_assert!` then it will work in a const function. We lose the debug print of the values but that isn't a big deal.

src/fastalloc/lru.rs
```diff
@@ -0,0 +1,331 @@
...
+            PReg::MAX + 1
+        ];
+        let no_of_regs = regs.len();
+        for i in 0..no_of_regs {
+            let (reg, prev_reg, next_reg) = (
+                regs[i],
+                regs[i.checked_sub(1).unwrap_or(no_of_regs - 1)],
+                regs[if i >= no_of_regs - 1 { 0 } else { i + 1 }],
+            );
+            data[reg.hw_enc()].prev = prev_reg.hw_enc() as u8;
```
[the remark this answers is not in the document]
d-sonuga: `hw_enc` is used because the LRUs are parted by regclass. A single LRU is for a single class.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1365 @@
...
+            if self.allocatable_regs.contains(preg) {
+                self.lrus[preg.class()].poke(preg);
+            }
+        }
+        for (_, op) in operands.fixed() {
+            let OperandConstraint::FixedReg(preg) = op.constraint() else {
+                unreachable!();
+            };
+            // Eviction has to be done separately to avoid using a fixed register
+            // as a scratch register.
```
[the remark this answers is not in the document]
d-sonuga: > Shouldn't remove_clobbers_from_available_pregs come before this? Otherwise a clobbered register might be used as a scratch register.


I don't think there is any problem with a clobbered register being used as a scratch register.

> In fact I think eviction for fixed operands should happen in process_operand_allocation. The same logic is needed no matter if the eviction is due to a fixed register or if the allocator selected a register with an existing vreg on it.

Eviction for fixed registers have to be done first because when iterating over the use and def operands are iterated over in any order. An example of why:

```
1. def v1 (reg), def v0 (fixed: p0)
2. use v1 (fixed: p0)
```

In the above example, v1 is already in p0 at inst 1. If we process v1 in inst 1 first, then v1 will be assigned to p0, even though it's not in the `available_pregs` set. The reason for this is the way `allocd_within_constraint` is implemented. In `allocd_within_constraint`, we can't just return false if the preg isn't in `available_pregs` because of the possibility of multiple operands using the same vreg. For example:

```
use v1 (fixed: p0), v1 (reg), v1 (reg)
```

The first encounter of v1 will remove p0 from `available_preg`. But for the remaining two, it will be wasteful to allocate new registers for them, so a check is made: `self.vreg_in_preg[preg.index()] == op.vreg()` to decide whether or not to conclude that it is allowed to be there. This introduces a new invariant: before `process_operand_allocation` is called, a vreg will be in a preg only if it's not reserved for some other vreg in the instruction. This necessitates the removal of all pregs from fixed registers before the use and def iterations.

src/fastalloc/lru.rs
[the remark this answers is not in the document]
Amanieu: Sure, but the index still fits in a `u8` and using the index avoids conversions between hw_enc and PReg (which just holds the index internally).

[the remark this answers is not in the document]
d-sonuga: The index also has the regclass in its lower bits which changes the value of the `hw_enc`

src/fastalloc/vregset.rs
```diff
@@ -1,104 +1,86 @@
...
 struct VRegNode {
-    next: u32,
-    prev: u32,
-    class: RegClass,
+    next: VRegIndex,
+    prev: VRegIndex,
+    vreg: VReg,
 }
 
 // Using a non-circular doubly linked list here for fast insertion,
```
Amanieu, line 15: This comment needs to be updated.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1365 @@
...
+    /// `inst_alloc_offsets[i]` is the offset into `allocs` for the allocations of
+    /// instruction `i`'s operands.
+    inst_alloc_offsets: Vec<u32>,
+}
+
+impl Allocs {
+    fn new<F: Function>(func: &F) -> (Self, u32) {
+        let operand_no_guess = func.num_insts() * 3;
+        let mut allocs = Vec::with_capacity(operand_no_guess);
+        let mut inst_alloc_offsets = Vec::with_capacity(operand_no_guess);
```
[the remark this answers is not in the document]
Amanieu: The size of `inst_alloc_offsets` should be the number of instructions, not the number of operands.

src/fastalloc/mod.rs
```diff
@@ -921,42 +921,6 @@ impl<'a, F: Function> Env<'a, F> {
                 self.vreg_in_preg[preg.index()] = VReg::invalid();
             }
         }
-        for (_, op) in operands.non_fixed_use() {
```
Amanieu: `non_fixed_use` can also be removed, it is no longer used.

src/fastalloc/vregset.rs
d-sonuga: Yeah,  I'm still getting to all these things. I'm not yet done making changes

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1365 @@
...
+        trace!(
+            "Block params at block {:?} beginning: {:?}",
+            block,
+            self.func.block_params(block)
+        );
+        trace!(
+            "Available pregs: {}",
+            self.available_pregs[OperandPos::Early]
+        );
+        let mut available_regs_for_scratch = self.available_pregs[OperandPos::Early];
```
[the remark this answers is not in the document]
d-sonuga: The issue here is that if that vreg is a livein or block param that's used in the first instruction, then it's possible for it to be overwritten if it's used as a scratch register.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1365 @@
...
+                        new_alloc,
+                        curr_alloc,
+                        op.class(),
+                        InstPosition::After,
+                    );
+                    // No need to set vreg_in_preg because it will be set during
+                    // `freealloc` if needed.
+                }
+                // Edits for use operands are added later to avoid inserting
+                // edits out of order.
```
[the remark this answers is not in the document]
Amanieu: If the vreg doesn't live past the instruction then `curr_alloc` is `None` and no move needs to be emitted.

I'm talking about a case like this:
```
1. use v1 (fixed: p1)
2. use v1 (fixed: p0)
```

In this case `curr_alloc` will be `p0` and `new_alloc` will be `p1`. A move needs to be inserted from `new_alloc` to `curr_alloc` after the instruction.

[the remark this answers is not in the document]
d-sonuga: If there are multiple operands using the same vreg in different allocations in the same instruction, then the moves will have to be inserted before:
```
1. use v1 (fixed: p1), use v1 (fixed: p2), use v1 (fixed: p3)
```
When the second v1 operand is encountered, `curr_alloc` will be p1 and inserting a move after will be incorrect since v1 doesn't live past this instruction.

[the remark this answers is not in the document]
Amanieu: Yes but that's a completely different kind of move that is handled in a different place (in the loop on line 948).

What I'm trying to say is that `if op.kind() == OperandKind::Def` on line 640 is incorrect, the moves need to be inserted for both Use and Def operands after the instruction. Multiple uses *also* need to have moves inserted before the instruction, but as I said before, that's handled elsewhere.

src/fastalloc/mod.rs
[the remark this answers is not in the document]
Amanieu: Doesn't this apply to all live registers in general? If a register is live then you can't use it as a scratch register.

However, if you look at it from another angle, you can treat all registers as available and just evict the vreg in a that register when you need to use it as a scratch register. The only difference is that evictions need to be inserted before the first instruction.

src/fastalloc/mod.rs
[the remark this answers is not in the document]
d-sonuga: When `curr_alloc` != `new_alloc` for use operands, it is always safe to insert the edit before the instruction. In some cases, it is safe to add the edit after, but not in all cases like when a vreg is used multiple times in an instruction and that vreg doesn't live past the instruction. For example:

```
1. def v0 (fixed: p1), use v1 (fixed: p1), use v1 (fixed: p2)
2. use v0 (fixed: p1)
```

In the above, when the second use operand is encountered in inst 1, v1's `curr_alloc` will be p1 and `new_alloc` will be p2. If we decide to insert an edit after, we'll have:


```
1. def v0 (fixed: p1), use v1 (fixed: p1), use v1 (fixed: p2)
move from p2 to p1
3. use v0 (fixed: p1)
```

Overwriting v0, resulting in an incorrect allocation. Checking for this special case of multiple operands in the same instruction will also be more expensive than simply checking if the operand is a def operand. The algorithm does not remember anything about whether or not the operand being allocated is the first or hundredth use of the same vreg.

On the other hand, if we decide to always insert edits for use operands before the instruction instead of after, there will be no incorrect allocations because it is always safe to insert the edits before the instruction for use operands. Remember that `vreg_allocs`, once initialized for a vreg, doesn't change unless the vreg is evicted or deallocated. So, if a use operand is expected to be in some allocation p0 after an instruction, like in:

```
1. use v1 (fixed: p1)
2. use v1 (fixed: p0)
```

When v1 is encountered in inst 2, `vreg_allocs` is set to p0 for it. When it's encountered in inst 1. `curr_alloc` will be p0 and `new_alloc` will be p1. `vreg_allocs` doesn't change and v1's `curr_alloc` remains p0. The loop afterwards that handles insertion of use operand edits will correctly insert the move from p0 to p1 before the instruction.

> What I'm trying to say is that if op.kind() == OperandKind::Def on line 640 is incorrect, the moves need to be inserted for both Use and Def operands after the instruction. Multiple uses also need to have moves inserted before the instruction, but as I said before, that's handled elsewhere.

No, it's not incorrect. Such a move definitely needs to be inserted after the instruction for a def, because the defined vreg doesn't live before the instruction. But it can always be safely inserted before the instruction for use operands. Can you give an example in which inserting before the instruction for a use operand will lead to an incorrect allocation?

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1365 @@
...
+                //
+                // When inst 1 is about to be processed, vreg_allocs[v1] will be p0.
+                // Suppose p1 is allocated to v0: this will create a fixed constraint for
+                // v1 and p1 will also be allocated to it.
+                // When it's time to process the v2 operand, vreg_allocs[v1] will still be p0
+                // because it doesn't change (except by an explicit fixed reg constraint which
+                // will not be a problem here) and it's possible for v2 to get p0 as an allocation,
+                // which is wrong. That will lead to the following scenario:
+                //
+                // move from p0 to p1   // Inserted due to reuse constraints
```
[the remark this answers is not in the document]
d-sonuga: I just realized that this loop actually has to be present, even before any evictions occur. An example:

```
1. def v0 (fixed: p23), use v1 (reg)
2. use v1 (fixed: p0)
// v0 is in stack_v0 and v1 is in stack_v1
```

If p23 is a stack slot, then we'll need to insert an edit to move from p23 to stack_v0 after inst 1 and a scratch register has to be used. Since p0 is still marked as available in `available_pregs`, it is possible for it to be used as scratch. If that happens then we'll have:

```
1. def v0 (fixed: p23), use v1 (reg)
move from stack_v1 to p0     // v1 is evicted
move from p23 to p0             // v1 is overwritten by v0
move from p0 to stack_v0
2. use v1 (fixed: p0)                // v0 is used instead of v1
```

Which is incorrect.

src/fastalloc/vregset.rs
```diff
@@ -0,0 +1,143 @@
...
+                };
+                num_vregs + 1
+            ],
+            head: VRegIndex::new(num_vregs),
+        }
+    }
+
+    pub fn insert(&mut self, vreg: VReg) {
+        // Intentionally assuming that the set doesn't already
+        // contain `vreg`.
```
cfallin, line 39: Can we assert this? It looks like we always use `self.items[vreg.vreg()]` as the node for this entry -- can we check if its `vreg` field is `VReg::invalid()`, for example, and set it as such when we remove?

src/fastalloc/vregset.rs
```diff
@@ -0,0 +1,143 @@
...
+impl VRegSet {
+    pub fn with_capacity(num_vregs: usize) -> Self {
+        Self {
+            items: vec![
+                VRegNode {
+                    prev: VRegIndex::new(num_vregs),
+                    next: VRegIndex::new(num_vregs),
+                    vreg: VReg::invalid()
+                };
+                num_vregs + 1
```
cfallin, line 31: it looks like we're using index 0 as a sentinel node (which is a great idea); do we need to add one below when we index into it with vreg numbers?

src/fastalloc/lru.rs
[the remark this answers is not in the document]
cfallin: Indeed, the issue is that this will result in sparse usage of `data` if we use the bitpacked index -- not the worst problem but something it'd be reasonable to avoid if we can. I'm personally fine with `hw_enc` here.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1391 @@
...
+        max_operand_len: u32,
+        num_insts: usize,
+        dedicated_scratch_regs: PartedByRegClass<Option<PReg>>,
+    ) -> Self {
+        // Some operands generate edits and some don't.
+        // The operands that generate edits add no more than two.
+        // Some edits are added due to clobbers, not operands.
+        // Anyways, I think this may be a reasonable guess.
+        let inst_edits_len_guess = max_operand_len as usize * 2;
+        let total_edits_len_guess = inst_edits_len_guess * num_insts;
```
cfallin, line 128: it might be good to collect some data manually: record `total_edits_len_guess` in this struct, and when done with allocation, print that guess, as well as `edits.len()`. This worst-case-size logic seems reasonable, but it also assumes all instructions will have `max_operand_len` (which may be large due to e.g. a call with many arguments) so I wonder if it may actually be much larger than needed. If we find that's the case, we could make a capacity estimate based on (say) `num_insts * K` where `K` is 2, or 4, or whatever factor we see.

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1391 @@
...
+impl Edits {
+    fn is_stack(&self, alloc: Allocation) -> bool {
+        if alloc.is_stack() {
+            return true;
+        }
+        if alloc.is_reg() {
+            return self.fixed_stack_slots.contains(alloc.as_reg().unwrap());
+        }
+        false
+    }
```
cfallin, line 147: We can write this as `alloc.is_stack() || (alloc.is_reg() && self.fixed_stack_slots.contains(...))` -- possibly a little more concise?

src/fastalloc/mod.rs
```diff
@@ -0,0 +1,1391 @@
...
+            // It is possible for an operand to have a fixed register constraint to
+            // a clobber.
+            OperandConstraint::FixedReg(preg) => alloc.is_reg() && alloc.as_reg().unwrap() == preg,
+            OperandConstraint::Reuse(_) => {
+                unreachable!()
+            }
+        }
+    }
+
+    fn evict_vreg_in_preg_before_inst(&mut self, inst: Inst, preg: PReg) {
```
cfallin, line 488: It looks like this function and `evict_vreg_in_preg` below are the same except for the `InstPosition` on the move, is that right? Can we pass it in as a parameter and deduplicate the two functions?

cfallin (commented):
I made another pass over this, and in `src/fastalloc/mod.rs` (the core algorithms), reviewed for plausibility and code style -- open threads with @Amanieu should be resolved still, but once they are, I'm happy to see this merge!

src/fastalloc/vregset.rs
d-sonuga: It's the last index that's used as a sentinel, so there isn't any need to add one

src/fastalloc/mod.rs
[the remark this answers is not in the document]
Amanieu: You are correct. You method is actually better than what I had in mind.

src/fastalloc/mod.rs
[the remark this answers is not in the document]
Amanieu: > I don't think there is any problem with a clobbered register being used as a scratch register.

You are correct, my bad.

> In the above example, v1 is already in p0 at inst 1. If we process v1 in inst 1 first, then v1 will be assigned to p0, even though it's not in the available_pregs set. The reason for this is the way allocd_within_constraint is implemented. In allocd_within_constraint, we can't just return false if the preg isn't in available_pregs because of the possibility of multiple operands using the same vreg. For example:

What I had in mind for such cases is to first check if there is already a use of this vreg in the same instruction and just reuse that allocation directly, skipping all the other checks. But your method also works, so it's fine to leave it as it is.

src/fastalloc/mod.rs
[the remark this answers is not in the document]
Amanieu: I think this still works because moves are emitted in reverse. Since the eviction of v1 happens first when a scratch register is selected, the actual move of stack_v1 to p0 will happen last.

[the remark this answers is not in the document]
d-sonuga: Right.

Amanieu (approved):

cfallin (approved):
Thanks a bunch for this new allocator implementation! Now that the initial version is complete and reviewed, let's merge it, and we can iterate on it and provide some performance summaries as followups.

d-sonuga:
A summary of performance:

Comparing the Rust compiler benchmarks with fastalloc vs ion, we have a speedup of 0.3%-18%.
Comparing with the Sightglass benchmarks: a few had no significant difference in performance. On the rest, compiling with fastalloc gives a speedup of 1.01-13x.
Profiling wasmtime compilation on inputs from the Sightglass benchmarks shows an increase in regalloc speed
of 3-20x.

PR #183, opened 2024-08-30
Remove safepoint support

elliottt:
As we have switched to modeling safepoints in clif directly, it's not longer necessary to track them in RA2. This PR is the first cut at removing safepoints from  RA2.

Co-authored-by: Nick Fitzgerald <fitzgen@gmail.com>


fitzgen:
@cfallin I believe this is ready for review now.

We are separately, as a follow up, also looking into removing `OperandConstraint::Stack`, as we believe it is now unused.

src/ion/moves.rs
```diff
@@ -563,11 +563,10 @@ impl<'a, F: Function> Env<'a, F> {
                     let inst = usedata.pos.inst();
                     let slot = usedata.slot;
                     let operand = usedata.operand;
-                    // Safepoints add virtual uses with no slots;
-                    // avoid these.
-                    if slot != SLOT_NONE {
-                        self.set_alloc(inst, slot as usize, alloc);
-                    }
+
+                    debug_assert_ne!(slot, SLOT_NONE);
```
cfallin, line 567: I suspect `SLOT_NONE` may now also be unused (the only point where I can see it used to create a `Use` is in the safepoint-handling loop in liveranges.rs, now deleted) -- can you remove if so?

cfallin (approved):
Really nice to see this code being deleted now that we don't need it -- thanks very much!

One potential further bit to remove below but otherwise LGTM.

src/ion/moves.rs
elliottt: Thanks for catching this!

PR #185, opened 2024-08-30
Remove the OperandConstraint::Stack variant

elliottt:
Following on from the removal of safepoints, remove the `OperandConstraint::Stack` and `Requirement::Stack` enum variants, as there's no longer any way to construct them.

Co-authored-by: Nick Fitzgerald <fitzgen@gmail.com>

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -216,14 +216,12 @@ impl LiveBundle {
         minimal: bool,
         fixed: bool,
         fixed_def: bool,
-        stack: bool,
     ) {
         debug_assert!(spill_weight <= BUNDLE_MAX_SPILL_WEIGHT);
         self.spill_weight_and_props = spill_weight
             | (if minimal { 1 << 31 } else { 0 })
             | (if fixed { 1 << 30 } else { 0 })
-            | (if fixed_def { 1 << 29 } else { 0 })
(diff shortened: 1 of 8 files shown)


src/ion/data_structures.rs
```diff
@@ -216,14 +216,12 @@ impl LiveBundle {
...
         fixed_def: bool,
-        stack: bool,
     ) {
         debug_assert!(spill_weight <= BUNDLE_MAX_SPILL_WEIGHT);
         self.spill_weight_and_props = spill_weight
             | (if minimal { 1 << 31 } else { 0 })
             | (if fixed { 1 << 30 } else { 0 })
-            | (if fixed_def { 1 << 29 } else { 0 })
-            | (if stack { 1 << 28 } else { 0 });
+            | (if fixed_def { 1 << 29 } else { 0 });
```
cfallin, line 224: Since we get bit 28 back here can we update `BUNDLE_MAX_SPILL_WEIGHT` as well (so the saturation point moves a little further out, leading to maybe slightly better alloc in deeply nested loops)?

cfallin (approved):
Nice, thank you!

src/ion/data_structures.rs
elliottt: Great idea!

PR #190, opened 2024-09-06
Stop moving Early FixedReg Use operands in certain conditions when Reuse operands are present.

Iizerd:
One of the reasons for the panic at [here](https://github.com/bytecodealliance/regalloc2/blob/main/src/ion/process.rs#L1253) is that `Early FixedReg Uses` are having their position updated to `Late` when there is already a `Late FixedReg Def` or clobber of the fixed `PReg`. This fixes that by not moving the operand's position if one of these already exist.

It's still possible to cause this behavior if a `Late Def` operand `Reuse`s an `Early FixedReg Def`. However that is also easy to fix if `Reuse` operands are checked to see if they reuse a `FixedReg` operand.

Thoughts? Am I missing something here? I ran the fuzzer overnight and compiled a few large binaries 200mb plus without fail; All run fine.

### DIFF
--- a/src/ion/liveranges.rs
+++ b/src/ion/liveranges.rs
@@ -594,7 +594,17 @@ impl<'a, F: Function> Env<'a, F> {
                                 if !reused_inputs.is_empty()
                                     && !reused_inputs.contains(&operand.vreg()) =>
                             {
-                                ProgPoint::after(inst)
+                                if let OperandConstraint::FixedReg(preg) = operand.constraint() {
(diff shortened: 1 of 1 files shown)


cfallin:
I think the new logic makes sense. However, as with #180, which I see is included here as the first commit, could you extend the fuzzer to generate multiple reuse-input defs, and fuzz with that?

Iizerd:
Yeah sorry about including #180 that again, I probably should have just put this commit in that same PR. This doesn't yet solve all the reason for that panic occurring, there is one more I'm working on.

PR #196, opened 2024-09-21
Adding new API that accepts resusable context.

jakubDoka:
Sadly due to how the code was structured, I needed to change the `Env' fields so basically everything that was used was changed as well. I did not benchmark anything yet (work in progress).

Context -> https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/Using.20context.20for.20.60TargetIsa.3A.3Acompile_function.60


jakubDoka:
I profiled the allocator a bit and I got rid of some hashmaps in favor of using the key as an index which improved things a little bit.

cfallin:
@jakubDoka could you summarize with a perf measurement (Sightglass or hyperfine of `wasmtime compile` is OK) the overall compile-time effect of this?

jakubDoka:
I had to run the bench separately because whatever I run last performs worse (if I run two identical programs, the first run always wins) (my CPU throttles)

```
[I] ::<~/p/r/wasmtime> hyperfine --runs 100 "./wasmtime.new compile -C parallel-compilation=n ../sightglass/benchmarks/pulldown-cmark/benchmark.wasm"
Benchmark 1: ./wasmtime.new compile -C parallel-compilation=n ../sightglass/benchmarks/pulldown-cmark/benchmark.wasm
  Time (mean ± σ):     136.7 ms ±   4.5 ms    [User: 128.8 ms, System: 7.1 ms]
  Range (min … max):   132.5 ms … 157.8 ms    100 runs

[I] ::<~/p/r/wasmtime> hyperfine --runs 100 "./wasmtime.old compile -C parallel-compilation=n ../sightglass/benchmarks/pulldown-cmark/benchmark.wasm"
Benchmark 1: ./wasmtime.old compile -C parallel-compilation=n ../sightglass/benchmarks/pulldown-cmark/benchmark.wasm
  Time (mean ± σ):     147.1 ms ±   4.6 ms    [User: 138.2 ms, System: 7.6 ms]
  Range (min … max):   140.1 ms … 163.2 ms    100 runs
```
Let me know If full perf stat could be useful
```
(spidermonkey)
15,471,712,064      cpu_core/cycles/u // wasmtime.old
14,390,147,910      cpu_core/cycles/u // wasmtime.new
(pulldown-cmark)
631,885,447      cpu_core/cycles/u // wasmtime.old
599,844,924      cpu_core/cycles/u // wasmtime.new
```

src/ion/process.rs
```diff
@@ -99,17 +101,18 @@ impl<'a, F: Function> Env<'a, F> {
 
             let mut skips = 0;
             'alloc: loop {
-                trace!("  -> PReg range {:?}", preg_range_iter.peek());
+                let Some(next) = preg_range_iter.as_slice().first() else {
```
cfallin, line 104: Why was this moved to the top of the loop? It looks like there's some additional reworking of the algorithm here but I'd like to avoid semantic changes like this when in a PR that adds allocator support.

src/ion/moves.rs
```diff
@@ -1001,6 +830,194 @@ impl<'a, F: Function> Env<'a, F> {
...
+    edits: Edits,
+    inter_block_sources: Vec<(VRegIndex, Allocation)>,
+    block_param_sources: FxHashMap<BlockparamSourceKey, Allocation>,
+    inter_block_dests: Vec<InterBlockDest>,
+    block_param_dests: Vec<BlockparamDest>,
+    reuse_input_insts: Vec<Inst>,
+}
+
+/// Buffered information about the previous liverange that was processed.
+struct PrevBuffer {
```
cfallin, line 853: Good idea moving this out of the function-local context, but can we have it in a separate PR (with code motion but no other changes) to make review easier / make any diffs apparent?

src/ion/merge.rs
```diff
@@ -170,14 +173,49 @@ impl<'a, F: Function> Env<'a, F> {
...
+                LiveRangeListEntry {
+                    range: CodeRange {
+                        from: ProgPoint::from_index(0),
+                        to: ProgPoint::from_index(0),
+                    },
+                    index: LiveRangeIndex::invalid(),
+                },
+            );
+
+            let mut reader = to_ranges.len() - from_list.len();
```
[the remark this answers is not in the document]
cfallin: On a little more thought after seeing some more semantics-altering changes below: I'd like to have this one separated into another PR so we can reason about it, see its impact, revert if necessary, etc separately from the mostly mechanical addition of allocator support in this PR.

src/ion/moves.rs
```diff
@@ -376,24 +205,16 @@ impl<'a, F: Function> Env<'a, F> {
...
-                        // register allocation.
-                        Entry::Occupied(mut entry) => {
-                            if !entry.get().is_reg() {
-                                entry.insert(alloc);
-                            }
-                        }
-                        Entry::Vacant(entry) => {
-                            entry.insert(alloc);
-                        }
+                    let entry = &mut inter_block_sources[block.index()];
```
cfallin, line 215: Can we keep the original code here to avoid non-mechanical changes (and do a separate PR if this change is worthwhile)?

src/ion/process.rs
```diff
@@ -1096,13 +1155,18 @@ impl<'a, F: Function> Env<'a, F> {
...
+                ) {
                     AllocRegResult::Allocated(alloc) => {
-                        self.stats.process_bundle_reg_success_any += 1;
+                        self.ctx.output.stats.process_bundle_reg_success_any += 1;
                         trace!(" -> allocated to any {:?}", preg_idx);
-                        self.spillsets[self.bundles[bundle].spillset].reg_hint =
+                        self.ctx.spillsets[self.ctx.bundles[bundle].spillset].reg_hint =
                             alloc.as_reg().unwrap();
-                        return Ok(());
+                        break 'o;
```
cfallin, line 1169: Can we choose a more descriptive name for this loop label (`'outer` or somesuch) and add a comment here (and below) where we replaced `return Ok(())` to indicate this is a successful allocation ("`// Success, return scratch memory to context and finish`" or similar)?

src/ion/data_structures.rs
```diff
@@ -13,19 +13,139 @@
...
+    pub fn clear(&mut self) {
+        self.values.clear();
+    }
+
+    pub fn contains_key(&self, key: &K) -> bool {
+        self.values.binary_search_by_key(key, |e| e.0).is_ok()
+    }
+
+    pub fn insert(&mut self, key: K, value: V) -> Option<V> {
+        // this is woth it becuase we often insert later liveranges
```
cfallin, line 107: s/woth/worth/ and capitalization and period on sentence.

src/ion/data_structures.rs
```diff
@@ -13,19 +13,139 @@
...
+        old_layout: core::alloc::Layout,
+        new_layout: core::alloc::Layout,
+    ) -> Result<core::ptr::NonNull<[u8]>, allocator_api2::alloc::AllocError> {
+        self.0.deref().shrink(ptr, old_layout, new_layout)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct BTreeMap<K, V> {
+    pub values: Vec<(K, V)>,
```
cfallin, line 94: I'm fairly worried about worst-case behavior here; to the extent that I don't know if I want to accept this replacement of the BTreeMap unless we can make a clear argument why we will never hit quadratic behavior. We've had fuzzbugs in the past that uncover quadratic cases and we have to protect against them (we consider it a DoS vector).

cfallin (commented):
Thanks for all this work! I'm happy to see the performance improvement. After reading through, I have a few high-level thoughts:

- I'd like to see some of the changes here pulled out into separate PRs. For example, you replace concatenate-and-sort steps with in-place merge when we know the liverange lists are already sorted; good optimization; but let's have that as a separate PR, so we can reason about this PR more easily and also so that we can test and revert those changes separately. Basically, I want this PR to be as close as possible to a mechanical replacement of data structures with arena-allocated or reused data structures, with no semantics changes. Some of the changes to the core allocation loop around the allocation-map iterator also make me very nervous.

- The replacement of the actual `BTreeMap` per physical reg with a sorted `Vec` makes me very nervous. It probably looks better on most benchmarks, but it has a quadratic worst-case (inserting in the middle), where a true btree does not; and we consider quadratic worst-cases bugs that we need to avoid. Is there an argument we can make about why this won't be seen in practice or...?

jakubDoka:
> The replacement of the actual BTreeMap per physical reg with a sorted Vec makes me very nervous. It probably looks better on most benchmarks, but it has a quadratic worst-case (inserting in the middle), where a true btree does not; and we consider quadratic worst-cases bugs that we need to avoid. Is there an argument we can make about why this won't be seen in practice or...?

I did not realize Wasmtime needs to deal with malicious code, good point, all of the changes here are guided by a simple principle: How to make the code do less. Using vec instead of BTree has many properties that CPUs like. I suspect that in `try_to_insert_bundle_to_reg`, replacing BTree with vec improves drastically because we mostly traverse the entries linearly for which Vec is more cache-friendly than BTree.

Now that I think about this, quadratic behavior is a problem if the function we compile is maliciously large, this is not the common case though, so what about doing a hybrid/adaptative approach? We could switch to BTreeMap dynamically if we detect Vec costing too much (it exceeds a certain size). What do you think @cfallin?

src/domtree.rs
```diff
@@ -12,24 +12,25 @@
...
 use alloc::vec::Vec;
 
-use crate::Block;
+use crate::{Block, VecExt};
 
 // Helper
 fn merge_sets(
     idom: &[Block], // map from Block to Block
-    block_to_rpo: &[Option<u32>],
+    block_to_rpo: &[u32],
```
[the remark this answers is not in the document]
jakubDoka: Good point, no performance wins detected

jakubDoka:
After reverting to BTree I see a regression of ~500m cycles.

jakubDoka:
After reverting this I see a regression of another ~250m cycles.

fitzgen:
> what about doing a hybrid/adaptative approach? We could switch to BTreeMap dynamically if we detect Vec costing too much (it exceeds a certain size).

Another, perhaps a little more out-there, idea is to make the whole algorithm generic over this container type, and make two instantiations: one with `Vec` and one with `BTreeSet`. Then, we try running the `Vec`-instantiated version first, but bail out if the length gets too long, at which point we fall back to the `BTreeSet`. If the length is a fixed size, we could even use an array instead of a `Vec`.

The failure checks and propagation overheads might dwarf any speed up though.

jakubDoka:
@cfallin are there any extra modifications needed? (considering I will add the other changes in a separate pr(s))

jakubDoka:
I realized there are some things I missed, so nvm

fuzz/Cargo.toml
```diff
@@ -60,3 +60,6 @@ doc = false
 debug = true
 debug-assertions = true
 overflow-checks = true
+#lto = true
```
cfallin, line 63: Can you say why these config options were added to `fuzz/Cargo.toml`?

Also, the commented-out line should be removed.

cfallin (commented):
Yes, please make sure all of my earlier comments were addressed; also one new one below from the start of my second pass.

fuzz/Cargo.toml
jakubDoka: The options are there to spend a bit more time optimizing the binary to hopefully get to run more fuzzing I think I forgot to remove this since it's not related to the topic

jakubDoka:
Okay, @cfallin, I double-checked everything, hopefully nothing was missed (I found one thing in the diff).

src/ion/liveranges.rs
```diff
@@ -476,7 +476,8 @@ impl<'a, F: Function> Env<'a, F> {
                 // register can be used multiple times in the same
                 // instruction is with an early-use and a late-def. Anything
                 // else is a user error.
-                let mut operand_rewrites: FxHashMap<usize, Operand> = FxHashMap::default();
+                let mut operand_rewrites = core::mem::take(&mut self.ctx.scratch_operand_rewrites);
```
cfallin, line 479: Do we need to do a `core::mem::take` on each instruction, and put it back below? Perhaps this could be hoisted out of the loop over instructions?

src/ion/merge.rs
```diff
@@ -132,7 +134,8 @@ impl<'a, F: Function> Env<'a, F> {
             // `to` bundle is empty -- just move the list over from
             // `from` and set `bundle` up-link on all ranges.
             trace!(" -> to bundle{} is empty; trivial merge", to.index());
-            let list = core::mem::replace(&mut self.bundles[from].ranges, smallvec![]);
+            let vc = LiveRangeList::new_in(self.ctx.bump());
```
cfallin, line 137: Rather than `vc`, could we call this something more descriptive like `empty_vec`? (And below?)

src/ion/mod.rs
```diff
@@ -106,38 +98,34 @@ impl<'a, F: Function> Env<'a, F> {
...
+        validate_ssa(func, &ctx.cfginfo)?;
     }
 
-    let mut env = Env::new(func, mach_env, cfginfo, enable_annotations);
+    ctx.annotations_enabled = enable_annotations;
+    let mut env = Env::new(func, mach_env, ctx);
     env.init()?;
 
-    let edits = env.run()?;
+    let mut moves = env.run()?;
```
cfallin, line 122: why is this renamed from `edits` to `moves`?

src/lib.rs
```diff
@@ -1606,3 +1636,93 @@ pub struct RegallocOptions {
...
+    }
+
+    fn preallocate(&mut self, cap: usize) -> &mut Self {
+        self.clear();
+        self.reserve(cap);
+        self
+    }
+}
+
+#[derive(Debug, Clone, Default)]
```
cfallin, line 1673: Doc-comment on this struct?

cfallin (commented):
Thanks! I have just a few more comments, but this is getting close.

deny.toml
```diff
@@ -1,7 +1,7 @@
 targets = [
     { triple = "x86_64-unknown-linux-gnu" },
     { triple = "x86_64-apple-darwin" },
-    { triple = "x86_64-pc-windows-msvc" },
+    { triple = "x86_64-pc-windows-msempty_vec" },
```
cfallin: I think this was an errant find-replace problem? `pc-windows-msempty_vec` sounds like an interesting platform, but not one that Rust supports...

src/lib.rs
```diff
@@ -1671,6 +1671,8 @@ impl<T> VecExt<T> for Vec<T> {
 }
 
 #[derive(Debug, Clone, Default)]
+/// Bump is a wrapper around `bumpalo::Bump` that can be cloned and also
+/// implements `Allocator`. Using this avoids lifetime polution of `Ctx`.
```
cfallin, line 1674: From the PoV of API docs consumers, there should probably be something here about what this is used for...

... actually, it seems it's not exposed at all in the `run_with_ctx` signature; can we make this `pub(crate)` then?

src/lib.rs
```diff
@@ -1580,13 +1591,32 @@ pub fn run<F: Function>(
...
     }
 }
 
+/// Run the allocator.
+pub fn run_with_ctx<F: Function>(
+    func: &F,
+    env: &MachineEnv,
+    options: &RegallocOptions,
+    ctx: &mut Ctx,
+) -> Result<(), RegAllocError> {
```
cfallin, line 1611: Could we return the `Output` here as with `run()`, by `mem::take`'ing it from the `Ctx`? I'd much prefer that to an implicit result in `ctx.output` -- OK to be imperative and implicit inside RA2 but a functional API is best.

cfallin (commented):
Thanks for your updates! A few more comments below; hopefully these are the last ones!

deny.toml
jakubDoka: yes and its outdated, I noticed it and fixed it

src/lib.rs
jakubDoka: It is implicit because the allocations inside the output are reused, of course, If we assume the user will return the output to the `Ctx` this is fine but IMO I'd rather hint in the doc where the output is located, that way you don't need to write more code to get optimal performance.

jakubDoka: Actually returning a reference could be a good middle ground

cfallin (approved):
OK, I think this is finally good to go. Thanks so much!

If you'd like, feel free to make a PR over in bytecodealliance/wasmtime to make use of `run_with_ctx` in Cranelift as well...

PR #197, opened 2024-09-29
Remove some final references to stack maps

fitzgen:
Stack maps are not provided by regalloc2 anymore. This removes the final references to stack maps in the codebase.


cfallin (approved):
Thanks!

Issue #194, opened 2024-09-11
Extend the fuzzer to check the debug locations output


d-sonuga:
The checker only checks for the correctness of the allocation dataflow but not for the debug locations output. Although the correctness of the debug locations output isn’t so critical, I think having it checked will still be a good idea, even if it’s just for testing.

cfallin:
I agree, this would be ideal to check. I don't have time to add it currently; if you want to make an attempt, I'm happy to review though.

d-sonuga:
Yeah, I’ll do it

Amanieu:
Note that debug info as currently implemented is only valid within the instructions themselves but not between original instructions. For example parallel move generation may evict a value to get a scratch register, but this isn't currently reflected in the debug info.

d-sonuga:
@cfallin, @Amanieu, is it correct for there to be multiple entries in the `debug_value_labels` input with the same label? Because the fuzzer is generating such inputs.

cfallin:
The same label can be attached to different vregs across different ranges of instruction indices, though these shouldn't overlap (i.e. the label should be in only one place at a time).

d-sonuga:
Right - so that means that this:
```
debug_value_labels: [
        (VReg(vreg = 0, class = Int), Inst(2), Inst(5), 53), 
        (VReg(vreg = 4, class = Int), Inst(2), Inst(8), 53), ]
```
isn't correct, right? Because the ranges overlap.

cfallin:
That indeed seems invalid! It looks like [this loop](https://github.com/bytecodealliance/regalloc2/blob/3c84b6fe1efc56a1a4c9a2a51e8a6daf0aa845e1/src/fuzzing/func.rs#L362-L375) ensures that over a single instance of the loop, it generates non-contiguous ranges, but it picks an arbitrary label and that label value might have been mentioned in another instance of the loop as well. I guess a fix could be to allocate a new label at that point (though this would prevent labels from moving between vregs, which is interesting to test, but doing better would require some sort of tracking where labels are free).

PR #200, opened 2024-10-18
Extend Fuzzer to Check Debug Locations

d-sonuga:
For #194.
The checker can now check whether the entries in the `debug_locations` output contain the vregs they're expected to contain as indicated by `Function::debug_value_labels`.
The checking doesn't happen by default; instead, it's only enabled during fuzzing.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -98,7 +98,7 @@
 use crate::{
     Allocation, AllocationKind, Block, Edit, Function, FxHashMap, FxHashSet, Inst, InstOrEdit,
     InstPosition, MachineEnv, Operand, OperandConstraint, OperandKind, OperandPos, Output, PReg,
-    PRegSet, VReg,
+    PRegSet, ProgPoint, VReg,
 };
 use alloc::vec::Vec;
 use alloc::{format, vec};
@@ -110,11 +110,11 @@ use smallvec::{smallvec, SmallVec};
 /// A set of errors detected by the regalloc checker.
 #[derive(Clone, Debug)]
 pub struct CheckerErrors {
-    errors: Vec<CheckerError>,
+    pub errors: Vec<CheckerError>,
 }
 
 /// A single error detected by the regalloc checker.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum CheckerError {
     MissingAllocation {
         inst: Inst,
@@ -166,6 +166,13 @@ pub enum CheckerError {
         into: Allocation,
         from: Allocation,
     },
+    ExpectedValueForDebug {
+        point: ProgPoint,
+        alloc: Allocation,
+        vreg: VReg,
+        found: CheckerValue,
+        label: u32,
+    },
 }
 
 /// Abstract state for an allocation.
@@ -174,7 +181,7 @@ pub enum CheckerError {
 /// universe-set as top and empty set as bottom lattice element. The
 /// meet-function is thus set intersection.
 #[derive(Clone, Debug, PartialEq, Eq)]
-enum CheckerValue {
+pub enum CheckerValue {
     /// The lattice top-value: this value could be equivalent to any
     /// vreg (i.e., the universe set).
     Universe,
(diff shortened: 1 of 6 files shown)


src/fastalloc/mod.rs
```diff
@@ -1209,11 +1209,35 @@ impl<'a, F: Function> Env<'a, F> {
...
 
+    fn point_range_intersect(
+        &self,
+        (start_inst, end_inst): (Inst, Inst),
+        (point_start, point_end): (ProgPoint, ProgPoint),
+    ) -> Option<(ProgPoint, ProgPoint)> {
+        if end_inst <= point_start.inst() || start_inst >= point_end.inst() {
+            None
+        } else {
+            let point0 = if point_start.inst() >= start_inst {
```
cfallin, line 1220: `ProgPoint` implements `Ord`, so could we use `std::cmp::min` / `std::cmp::max` here?

src/fastalloc/tests.rs
```diff
@@ -102,10 +103,118 @@ fn test_debug_locations2() {
...
+    }]);
+    f.debug_value_labels = vec![
+        (v(0), i(0), i(4), 32),
+        (v(2), i(2), i(4), 70),
+        (v(2), i(2), i(4), 71),
+        (v(3), i(3), i(4), 34),
+    ];
+    let mut result = run(&f, &mach_env, &options).unwrap();
+    /*
+    The correct debug_locations output
```
cfallin, line 142: Are we waiting for a fix before asserting this output instead? If so could we add a note here that this is to be asserted once we fix [thing]?

src/checker.rs
```diff
@@ -690,6 +697,128 @@ pub(crate) enum CheckerInst {
...
+
+impl DebugLocations {
+    fn ranges_overlaps(
+        (start_point, end_point): (ProgPoint, ProgPoint),
+        (start_inst, end_inst): (Inst, Inst),
+    ) -> Option<(ProgPoint, ProgPoint)> {
+        if end_inst <= start_point.inst() || start_inst >= end_point.inst() {
+            None
+        } else {
+            let point0 = if start_point.inst() >= start_inst {
```
cfallin, line 720: Similar here to below -- `std::cmp::min` / `std::cmp::max`?

cfallin: And can we name this `intersects`, and factor it out to a common method on `ProgPoint` shared by the other use below? I see the args are slightly different -- ProgPoint/Inst here, Inst/ProgPoint below, but unless there are subtleties around the handling of begin/end, we could take ProgPoints for both ranges and convert Insts to ProgPoints as needed at callsites I think?

src/checker.rs
```diff
@@ -690,6 +697,128 @@ pub(crate) enum CheckerInst {
...
+                ProgPoint::before(end_inst)
+            };
+            Some((point0, point1))
+        }
+    }
+
+    fn new<F: Function>(f: &F, output: &Output) -> Self {
+        let mut expected_vreg_locations = FxHashMap::default();
+        for (label, start_point, end_point, alloc) in &output.debug_locations {
+            for (vreg, start_inst, end_inst, in_label) in f.debug_value_labels() {
```
cfallin, line 737: I'm a bit worried that this could result in quadratic runtime during fuzzing (and that the fuzzer will discover and exercise this worst case) -- could we optimize this somehow by building a map? For example we could have a label -> [array of allocs] dense-map, where [array of allocs] is indexed by the ProgPoint's index; then when processing `debug_value_labels` we can cross-reference that.

src/checker.rs
```diff
@@ -690,6 +697,128 @@ pub(crate) enum CheckerInst {
...
+        // for moving the vreg into the expected allocation.
+        if inst == end_inst && end_point.pos() == InstPosition::After {
+            return (true, false);
+        }
+        (false, false)
+    }
+
+    fn entries_covering(&self, inst: Inst) -> Vec<(bool, bool, DebugLocationEntry)> {
+        let mut entries = vec![];
+        for entry in self.expected_vreg_locations.keys().copied() {
```
cfallin, line 778: Here is another place I'm a bit worried about checker-time blowup: we're iterating over all vreg locations (which is O(|num vregs| * |program size|)) once per `inst`, so if |vregs| ~= |insts|, this is approximately cubic overall, right?

cfallin (commented):
Thanks for working on this! I have a few thoughts below.

My main thoughts are around algorithmic complexity -- it seems we're scanning the join of the allocation results, the debug-labels list, and the program in various combinations that could result in quadratic or cubic checking cost, in a way that might significantly slow down fuzzing.

At a high level I think what we might want to do is to build searchable maps of both the value-labels input, and the debug-locations output, and then whenever a new vreg is defined in a physical register, check if the vreg carries a label and if so check that the label appears in the debug data, or something like that. What do you think?

d-sonuga:
@cfallin, I'm definitely aware that this is inefficient: I actually just wanted some thoughts on whether or not the correctness of the debug locations is actually being checked reasonably.

My bad for not being explicit about that in the initial comment :sweat_smile:.

d-sonuga:
I'd prefer to leave it open as a draft. I still intend to work on this.

PR #204, opened 2024-12-02
Improve performance by optimizing  bundle merging logic.

### LINKED ISSUE #203: Performance regression with rustc 1.81
rustc 1.81 completely replaced the implementation of [slice::sort_unstable_by_key](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable_by_key). These changes are mentioned in the[ release notes](https://releases.rs/docs/1.81.0/#libraries), and more information can be found in the corresponding PR: https://github.com/rust-lang/rust/pull/124032/.

The new sorting algorithm was supposed to be faster in the general case, but in some specific cases it looks like it is way slower, and this has impacted the `wasmtime` crate via `regalloc2`.  I've experienced a 10x times slowdown when compiling some WASM modules using `wasmtime` with rustc 1.81 (more details can be found in: https://github.com/VirusTotal/yara-x/issues/209)

I tracked down the issue to be caused by the [merge_bundles](https://github.com/bytecodealliance/regalloc2/blob/e684ee565d5b55e5d7bde6efa50996d5505f038c/src/ion/merge.rs#L24) function, which uses `sort_unstable_by_key`. According to my investigation, `sort_unstable_by_key` is being called with slices that are mostly sorted. Apparently, two already sorted slices are being concatenated and then sorted again. This is confirmed by the following comment in the code of `merge_bundles`:

```
// Two non-empty lists of LiveRanges: concatenate and
// sort. This is faster than a mergesort-like merge into a new
// list, empirically.
```

This pattern, which worked well with the previous implementation of `sort_unstable_by_key`, seems to be pathologically bad for the new one. The assumption that concatenating the two lists and sorting them again is faster than merging the two sorted lists likely no longer holds true.

If you are curious about which WASM code is triggering this issue, it is an implementation of the "switch" statement in WASM using multiple nested blocks and a `br_table` instruction that jumps out the block that corresponds to the input value. The overall schema is shown below.

```wasm
(func $switch_example (param $x i32)
  (block $default  ;; Default branch
    (block $case2   ;; Case 2
      (block $case1 ;; Case 1
        (block $case0 ;; Case 0
          ;; Use br_table to jump based on $x
          (local.get $x)    ;; Push x onto the stack
          (br_table $case0 $case1 $case2 $default) ;; Jump to the right case
        )
        ;; Code for Case 0
        ;; Add your case 0 logic here
        (br $default) ;; Continue to the end
      )
      ;; Code for Case 1
      ;; Add your case 1 logic here
      (br $default) ;; Continue to the end
    )
    ;; Code for Case 2
    ;; Add your case 2 logic here
    (br $default) ;; Continue to the end
  )
  ;; Default case code
)
```

plusvic:
While merging two bundles, the `merge_bundles` function appends two sorted vectors and sort the resulting vector again. This approach used to be fast in most cases, but rustc 1.81 introduced changes in sorting algorithm that made `sort_unstable_by_key` to behave very bad with vectors that are almost sorted.

This introduces an optimization consisting in handling the special case where the vector being appended contains a single item differently. This case is very common, and there's a benefit in handling it differently both with rust 1.81 and with earlier versions.

Fixes #203

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,3 +1,4 @@
+[graph]
 targets = [
     { triple = "x86_64-unknown-linux-gnu" },
     { triple = "x86_64-apple-darwin" },
@@ -7,8 +8,6 @@ targets = [
 
 # https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html
 [advisories]
-vulnerability = "deny"
-unmaintained = "deny"
 yanked = "deny"
 ignore = []
 
@@ -19,6 +18,7 @@ allow = [
     "Apache-2.0",
     "MIT",
     "Unicode-DFS-2016",
+    "Unicode-3.0",
 ]
 
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
--- a/src/ion/merge.rs
+++ b/src/ion/merge.rs
@@ -170,19 +170,33 @@ impl<'a, F: Function> Env<'a, F> {
             ranges_to
         );
 
-        // Two non-empty lists of LiveRanges: concatenate and
-        // sort. This is faster than a mergesort-like merge into a new
-        // list, empirically.
         let empty_vec = LiveRangeList::new_in(self.ctx.bump());
-        let from_list = core::mem::replace(&mut self.bundles[from].ranges, empty_vec);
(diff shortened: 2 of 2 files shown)


cfallin:
(I'm looking into the cargo-deny failure -- seems to be a license issue with a crate dependency)

PR #207, opened 2024-12-26
Fix broken link in README (NFC)

c8ef:


README.md
```diff
@@ -9,7 +9,7 @@ In addition, it contains substantial amounts of testing infrastructure
 (fuzzing harnesses and checkers) that does not exist in the original
 IonMonkey allocator.
 
-See the [design overview](doc/DESIGN.md) for (much!) more detail on
+See the [design overview](doc/ION.md) for (much!) more detail on
```
cfallin, line 12: Can you link to the toplevel `GENERAL.md` instead?

cfallin (commented):
Thanks!

README.md
cfallin: (Or maybe both -- the content that was originally referred to here is split now)

cfallin (approved):

PR #208, opened 2025-01-13
refactor: use `&'static [PReg]` instead of `Vec<PReg>`

JonasKruckenberg:
This PR makes `MachineEnv` use `&'static [PReg]` instead of `Vec<PReg>` to make it possible to place `MachineEnv` into `static`s as part of https://github.com/bytecodealliance/wasmtime/pull/8489.

related: #177.

### DIFF
--- a/src/fastalloc/mod.rs
+++ b/src/fastalloc/mod.rs
@@ -245,9 +245,9 @@ impl<'a, F: Function> Env<'a, F> {
     fn new(func: &'a F, env: &'a MachineEnv) -> Self {
         use alloc::vec;
         let mut regs = [
-            env.preferred_regs_by_class[RegClass::Int as usize].clone(),
-            env.preferred_regs_by_class[RegClass::Float as usize].clone(),
-            env.preferred_regs_by_class[RegClass::Vector as usize].clone(),
+            Vec::from(env.preferred_regs_by_class[RegClass::Int as usize]),
+            Vec::from(env.preferred_regs_by_class[RegClass::Float as usize]),
+            Vec::from(env.preferred_regs_by_class[RegClass::Vector as usize]),
         ];
         regs[0].extend(
             env.non_preferred_regs_by_class[RegClass::Int as usize]
--- a/src/fastalloc/tests.rs
+++ b/src/fastalloc/tests.rs
@@ -143,15 +143,17 @@ impl RealFunction {
 fn mach_env(no_of_regs: usize) -> MachineEnv {
     MachineEnv {
         preferred_regs_by_class: [
-            (0..no_of_regs)
-                .map(|no| PReg::new(no, RegClass::Int))
-                .collect(),
-            vec![],
-            vec![],
+            Vec::leak(
+                (0..no_of_regs)
+                    .map(|no| PReg::new(no, RegClass::Int))
+                    .collect(),
+            ),
+            &[],
+            &[],
         ],
-        non_preferred_regs_by_class: [vec![], vec![], vec![]],
+        non_preferred_regs_by_class: [&[], &[], &[]],
         scratch_by_class: [None, None, None],
-        fixed_stack_slots: vec![],
+        fixed_stack_slots: &[],
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -338,13 +338,13 @@ impl From<&MachineEnv> for PRegSet {
     fn from(env: &MachineEnv) -> Self {
         let mut res = Self::default();
 
-        for class in env.preferred_regs_by_class.iter() {
+        for class in env.preferred_regs_by_class {
             for preg in class {
                 res.add(*preg)
             }
         }
 
-        for class in env.non_preferred_regs_by_class.iter() {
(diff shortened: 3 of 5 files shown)


src/fastalloc/mod.rs
```diff
@@ -245,9 +245,9 @@ impl<'a, F: Function> Env<'a, F> {
     fn new(func: &'a F, env: &'a MachineEnv) -> Self {
         use alloc::vec;
         let mut regs = [
-            env.preferred_regs_by_class[RegClass::Int as usize].clone(),
-            env.preferred_regs_by_class[RegClass::Float as usize].clone(),
-            env.preferred_regs_by_class[RegClass::Vector as usize].clone(),
+            Vec::from(env.preferred_regs_by_class[RegClass::Int as usize]),
```
JonasKruckenberg, line 248: don't love this, open for better ideas

bjorn3: Maybe use `.to_owned()` instead of `Vec::from`?

src/lib.rs
```diff
@@ -1434,6 +1434,12 @@ impl<'a> Iterator for OutputIter<'a> {
 /// are available to allocate and what register may be used as a
 /// scratch register for each class, and some other miscellaneous info
 /// as well.
+///
+/// Note that `MachineEnv` is designed to be a global configuration struct that programs
+/// will have very few of and generally want to keep around for their entire lifetime.
+/// In order to make static initialization easier the registers lists are static slices instead
+/// of `Vec`s. If your use case depends on dynamically creating the registers lists, consider
+/// `[Vec::leak`].
```
bjorn3, line 1442:
```suggestion
/// [`Vec::leak`].
```

src/fastalloc/mod.rs
JonasKruckenberg: hmm, yeah doesn't really change anything about it but might look nicer

JonasKruckenberg:
Hmm seems libfuzz doesn't like `Vec::leak`?

JonasKruckenberg:
> Hmm seems libfuzz doesn't like `Vec::leak`?

interestingly I locally can't replicate the CI fail

src/lib.rs
```diff
@@ -1434,15 +1434,20 @@ impl<'a> Iterator for OutputIter<'a> {
 /// are available to allocate and what register may be used as a
 /// scratch register for each class, and some other miscellaneous info
 /// as well.
+///
+/// Note that `MachineEnv` is designed to be a global configuration struct that programs
+/// will have very few of and generally want to keep around for their entire lifetime.
+/// In order to make static initialization easier the registers lists are static slices instead
+/// of `Vec`s. If your use case depends on dynamically creating the registers lists, consider
+/// [`Vec::leak`].
```
cfallin, line 1442: I'd prefer not to give this suggestion in the doc-comment: since it's parameterized on a lifetime (rather than hardcoding `'static`), it's possible to build a `MachineEnv` dynamically that is a borrowed view on owned `Vec`s, basically giving the same capabilities as before. This is fairly idiomatic in Rust, so I don't think we need a special note or recommendation.

src/fuzzing/func.rs
```diff
@@ -624,30 +624,32 @@ impl core::fmt::Debug for Func {
...
-    let fixed_stack_slots = (32..63)
-        .flat_map(|i| {
-            [
-                PReg::new(i, RegClass::Int),
-                PReg::new(i, RegClass::Float),
-                PReg::new(i, RegClass::Vector),
-            ]
-        })
-        .collect();
+    let fixed_stack_slots = Vec::leak(
```
cfallin, line 642: I think we need to avoid `Vec::leak` here: fuzzing will continually generate new functions with machine envs, and we cannot allocate indefinitely. Can you build a static initializer, since the range (`32..63`) is static?

src/fastalloc/tests.rs
```diff
@@ -140,18 +140,20 @@ impl RealFunction {
...
-fn mach_env(no_of_regs: usize) -> MachineEnv {
+fn mach_env(no_of_regs: usize) -> MachineEnv<'static> {
     MachineEnv {
         preferred_regs_by_class: [
-            (0..no_of_regs)
-                .map(|no| PReg::new(no, RegClass::Int))
-                .collect(),
-            vec![],
-            vec![],
+            Vec::leak(
```
cfallin, line 146: We definitely can't leak while fuzzing -- can we instead have a few static `MachineEnv`s in scope, and pick one of them? We don't necessarily need to fuzz with arbitrary `no_of_regs`, just with a few values that are interesting -- the two extremes of its range and a few in the middle, for example.

cfallin (commented):
A few comments below -- thanks for working on this! I think the general shape is right, I just have some thoughts on the way the fuzzing testcase generation is working (we need to avoid leaks).

src/lib.rs
JonasKruckenberg: ah yeah good catch, that was from before I allowed any lifetimes. This notice doesn't make much sense anymore

PR #211, opened 2025-03-25
Panics on non consecutive blocks and vregs ids

iwanders:
Hi, currently it is possible for two panics to happen when a function's SSA is being validated:
- If the block's numbering does not start at `0` or is non consecutive.
- If the vreg's numbering does not start at `0` or is non consecutive.

I ran into it after trying to use regalloc2's allocators in my toy/educational compiler project... I ended up with a hacked up clif file, something like:
```
test run
test compile
target x86_64

function %branch_diverge(i64, i64) -> i64 system_v {

block1(v0: i64, v1: i64):
    brif v0, block2, block2

block2:
    v6 = iconst.i64 1
    v11 = iadd v6, v1
    return v11
}
; run: %branch_diverge(0, 0) == 1
```
Which worked fine in my register allocator, as it uses hashmaps, but when trying to run regalloc2's allocators I got panics. Running `clif-util test` from cranelift passes on this file, that probably relabels all ids before passing them to the register allocator?.

I'm not sure if non conseuctive ids violate an unwritten rule about SSA notation, but I didn't find anything in the documentation about this and if it is not allowed, ideally the ssa validation should catch it and report an error, such that users can easily understand what the problem is.

The first one happens in the [second argument](https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/fastalloc/mod.rs#L1290) of the `validate_ssa` call. Specifically here:
https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/postorder.rs#L38
in the case the index of a block is outside of `num_blocks`.

The other one here:
https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/ssa.rs#L19
In case the vregs ids are not consecutive.

A standalone minimal example that demonstrates the two problems can be found [here](https://gist.github.com/iwanders/433881a4d601e6e7d89fb817d146a0c6).

<details>
<summary>Backtraces from that example</summary>

```
thread 'main' panicked at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/postorder.rs:38:17:
index out of bounds: the len is 2 but the index is 2
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_bounds_check
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:273:5
   3: regalloc2::postorder::calculate
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/postorder.rs:38:17
   4: regalloc2::cfg::CFGInfo::init
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/cfg.rs:53:9
   5: regalloc2::cfg::CFGInfo::new
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/cfg.rs:46:9
   6: regalloc2::fastalloc::run
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/fastalloc/mod.rs:1290:29
   7: regalloc2::run
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/lib.rs:1600:13
   8: regalloc_panic::main
             at ./grus_codegen/examples/regalloc_panic.rs:173:21
   9: core::ops::function::FnOnce::call_once
             at /home/ivor/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
```

```
thread 'main' panicked at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/ssa.rs:19:26:
index out of bounds: the len is 4 but the index is 6
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_bounds_check
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:273:5
   3: <usize as core::slice::index::SliceIndex<[T]>>::index
             at /home/ivor/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:274:10
   4: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /home/ivor/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:16:9
   5: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /home/ivor/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:3361:9
   6: regalloc2::ssa::validate_ssa::{{closure}}
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/ssa.rs:19:26
   7: regalloc2::ssa::validate_ssa
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/ssa.rs:33:21
   8: regalloc2::fastalloc::run
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/fastalloc/mod.rs:1290:9
   9: regalloc2::run
             at /home/ivor/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.11.1/src/lib.rs:1600:13
  10: regalloc_panic::main
             at ./grus_codegen/examples/regalloc_panic.rs:192:22
  11: core::ops::function::FnOnce::call_once
             at /home/ivor/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5

```

</details>

Commits are split out:
- c536d39210aced8de6d5bdf757f64191452fe3bb adds checks that mitigate the panics.
- a3f088231e7a0d4d859709d80c3588f3e7fbb45a adds documentation to the `RegAllocError` enum, and a note in `Function` that states the requirement on the indices.
- cf441a96feb03dd5ccd6b65b7df972669f085774 adds some drive-by fixes to the documentation, they gave warnings about `[alloc_s]` not being an existing token, and `Operand::fixed` not existing, I think that should be `OperandConstraint::FixedReg` now?, lastly `target/` in the `.gitignore` doesn't match a symlink (I always use one to a ramdisk).

With these changes, the mentioned minimal non working example runs cleanly en reports `Err` as a return from the `run` method for both allocators.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -912,7 +912,7 @@ impl Operand {
     }
 
     /// If this operand is for a fixed non-allocatable register (see
-    /// [`OperandConstraint::FixedReg`]), then returns the physical register that it will
+    /// [`Operand::fixed_nonallocatable`]), then returns the physical register that it will
     /// be assigned to.
     #[inline(always)]
     pub fn as_fixed_nonallocatable(self) -> Option<PReg> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -912,7 +912,7 @@ impl Operand {
     }
 
     /// If this operand is for a fixed non-allocatable register (see
-    /// [`Operand::fixed`]), then returns the physical register that it will
+    /// [`OperandConstraint::FixedReg`]), then returns the physical register that it will
     /// be assigned to.
     #[inline(always)]
     pub fn as_fixed_nonallocatable(self) -> Option<PReg> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1120,6 +1120,9 @@ pub enum AllocationKind {
 ///
 /// (This trait's design is inspired by, and derives heavily from, the
 /// trait of the same name in regalloc.rs.)
+///
+/// The ids used in [`Block`] and [`VReg`] must start their numbering
+/// at zero and be sequential.
 pub trait Function {
     // -------------
     // CFG traversal
@@ -1557,10 +1560,12 @@ pub enum RegAllocError {
     /// a block param.
     SSA(VReg, Inst),
     /// Invalid basic block: does not end in branch/ret, or contains a
-    /// branch/ret in the middle.
+    /// branch/ret in the middle, or the VReg ids do not start at zero
+    /// or aren't numbered sequentially.
     BB(Block),
     /// Invalid branch: operand count does not match sum of block
-    /// params of successor blocks.
+    /// params of successor blocks, or the block ids do not start at
+    /// zero or aren't numbered sequentially.
     Branch(Inst),
     /// A VReg is live-in on entry; this is not allowed.
     EntryLivein,
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 Cargo.lock
-target/
+target
 .*.swp
 *~
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -56,7 +56,7 @@ impl CFGInfo {
             &mut ctx.visited,
             &mut self.postorder,
             |block| f.block_succs(block),
-        );
+        )?;
 
         domtree::calculate(
             nb,
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -65,12 +65,12 @@
 //!
 //!   - `Edit::Move` inserted by RA:       [ alloc_d := alloc_s ]
 //!
-//!       A' = A[alloc_d → A[alloc_s]]
+//!       A' = A[alloc_d → A\[alloc_s\]]
 //!
 //!   - statement in pre-regalloc function [ V_i := op V_j, V_k, ... ]
 //!     with allocated form                [ A_i := op A_j, A_k, ... ]
 //!
-//!       A' = { A_k → A[A_k] \ { V_i } for k ≠ i } ∪
+//!       A' = { A_k → A\[A_k\] \ { V_i } for k ≠ i } ∪
 //!            { A_i -> { V_i } }
 //!
 //!     In other words, a statement, even after allocation, generates
@@ -81,7 +81,7 @@
 //!   - Parallel moves or blockparam-assignments in original program
 //!                                       [ V_d1 := V_s1, V_d2 := V_s2, ... ]
 //!
-//!       A' = { A_k → subst(A[A_k]) for all k }
+//!       A' = { A_k → subst(A\[A_k\]) for all k }
 //!            where subst(S) removes symbols for overwritten virtual
 //!            registers (V_d1 .. V_dn) and then adds V_di whenever
 //!            V_si appeared prior to the removals.
--- a/src/fuzzing/func.rs
+++ b/src/fuzzing/func.rs
@@ -189,7 +189,7 @@ impl FuncBuilder {
 
     fn compute_doms(&mut self) {
         let f = &self.f;
-        postorder::calculate(
+        let _ = postorder::calculate(
             self.f.blocks.len(),
             Block::new(0),
             &mut vec![],
--- a/src/postorder.rs
+++ b/src/postorder.rs
@@ -5,7 +5,7 @@
 
 //! Fast postorder computation.
 
-use crate::{Block, VecExt};
+use crate::{Block, RegAllocError, VecExt};
 use alloc::vec::Vec;
 use smallvec::{smallvec, SmallVec};
 
@@ -15,7 +15,7 @@ pub fn calculate<'a, SuccFn: Fn(Block) -> &'a [Block]>(
     visited_scratch: &mut Vec<bool>,
     out: &mut Vec<Block>,
     succ_blocks: SuccFn,
-) {
+) -> Result<(), RegAllocError> {
     // State: visited-block map, and explicit DFS stack.
     struct State<'a> {
         block: Block,
@@ -26,7 +26,10 @@ pub fn calculate<'a, SuccFn: Fn(Block) -> &'a [Block]>(
     let mut stack: SmallVec<[State; 64]> = smallvec![];
     out.clear();
 
-    visited[entry.index()] = true;
+    let entry_visit = visited
+        .get_mut(entry.index())
+        .ok_or(RegAllocError::BB(entry))?;
+    *entry_visit = true;
     stack.push(State {
         block: entry,
         succs: succ_blocks(entry).iter(),
@@ -35,8 +38,11 @@ pub fn calculate<'a, SuccFn: Fn(Block) -> &'a [Block]>(
     while let Some(ref mut state) = stack.last_mut() {
         // Perform one action: push to new succ, skip an already-visited succ, or pop.
         if let Some(&succ) = state.succs.next() {
-            if !visited[succ.index()] {
-                visited[succ.index()] = true;
+            let succ_visit = visited
+                .get_mut(succ.index())
+                .ok_or(RegAllocError::BB(succ))?;
+            if !*succ_visit {
+                *succ_visit = true;
                 stack.push(State {
                     block: succ,
                     succs: succ_blocks(succ).iter(),
@@ -47,4 +53,5 @@ pub fn calculate<'a, SuccFn: Fn(Block) -> &'a [Block]>(
             stack.pop();
         }
     }
+    Ok(())
 }
--- a/src/ssa.rs
+++ b/src/ssa.rs
@@ -16,6 +16,10 @@ pub fn validate_ssa<F: Function>(f: &F, cfginfo: &CFGInfo) -> Result<(), RegAllo
     for block in 0..f.num_blocks() {
         let block = Block::new(block);
         let mut def = |vreg: VReg, inst| {
+            if vreg.vreg() >= defined_in.len() {
+                trace!("VRegs not numbered consecutively {:?}", vreg);
+                return Err(RegAllocError::SSA(vreg, inst));
+            }
             if defined_in[vreg.vreg()].is_valid() {
                 trace!("Multiple def constraints for {:?}", vreg);
                 Err(RegAllocError::SSA(vreg, inst))


cfallin:
Hi @iwanders  -- thanks for the report. 

Actually I think the issue is a bit different, and much simpler, than this PR implies: regalloc2 requires accurate information about the range of block IDs, instruction IDs, and VReg IDs, because it pre-allocates structures indexed by these IDs. If you provide inaccurate information, the code will panic.

Non-contiguous ranges, on the other hand, are perfectly fine; the allocator does not assume that the index spaces are fully filled.

The `num_vregs`, `num_blocks` and `num_insts` methods are perhaps ambiguously named: in the case of a sparse index space, these are not really the number of used indices, but the total size of the index space. (In other words, max-index plus one.) What I had meant by the names was exactly that -- "number of indices" rather than "number of things attached to these indices".

So I don't think this PR is the right solution, but we should properly document this, and we should validate that all IDs are in range so we get a clearer error. Would you be willing to do that?

iwanders:
> because it pre-allocates structures indexed by these IDs. If you provide inaccurate information, the code will panic.
> The num_vregs, num_blocks and num_insts methods are perhaps ambiguously named: in the case of a sparse index space, these are not really the number of used indices, but the total size of the index space.

Yeah, I think this is the main thing... the [docs](https://docs.rs/regalloc2/latest/regalloc2/trait.Function.html#tymethod.num_blocks) for those functions currently states they should return the number of blocks, not the value of the index of the highest block (plus one). So to make the code work, the only logical conclusion for me was to require the indices to be sequential and start at zero, otherwise they don't all end up with an index below `num_blocks`. Anyways, I get what you mean :+1: 

> So I don't think this PR is the right solution, but we should properly document this, and we should validate that all IDs are in range so we get a clearer error. Would you be willing to do that?

Yeah, no problem, I'll update it tomorrow. The fix (c536d39210aced8de6d5bdf757f64191452fe3bb) only needs a rewording of the trace on line 20, but the actual checks only check whether it's in bounds of the allocated container. As for the docs of the `num_blocks` function; we can just have the documentation explain that it's the number of elements in pre-allocated containers internally, so the user should pass the highest block id plus one, not the actual number of blocks used in the function. That explanation makes it clear why it is named like this.

iwanders:
Okay, I pushed a commit that rewords the documentation and makes the checks check the length. 

> Non-contiguous ranges, on the other hand, are perfectly fine; the allocator does not assume that the index spaces are fully filled.

About this though, there's a loop over the possible block indices here:

https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/cfg.rs#L79-L83

And the signature for `block_insn` does not return an optional:
https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/lib.rs#L1138

The naive solution in that case is to return an empty range, from instruction 0 to instruction 0, but that results in a debug assert:
https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/index.rs#L149

So, to handle non-consecutive blocks, one would have to do something like this:
```rust
    fn block_insns(&self, block: regalloc2::Block) -> InstRange {
        println!("block: {block:?}");
        if let Some(r) = self.blocks.get(&block) {
            self.blocks[&block]
        } else {
            println!("Don't have this block, returning a dummy range");
            InstRange::new(Inst::new(0), Inst::new(1))
        }
    }
```
which is less than intuitive as a workaroung?

I wonder if, instead of `0..f.num_blocks()`, we should do a traversal of the blocks through the entry block and block successors?

cfallin:
Doesn't that assert in `InstRange::new()` permit a range of `0..0` as it's a `<=` comparison?

I suppose you're finding a few places where a contiguous-indices assumption has leaked in, which makes me think that maybe the most prudent solution is just to check that property. It's certainly what Cranelift provides in its input, so that will be the path of least resistance for other users as well.

iwanders:
> Doesn't that assert in InstRange::new() permit a range of 0..0 as it's a <= comparison?

Sorry, yes, we hit the assert one function below that assert on 155;

https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/index.rs#L155

Which is called from the `CFGInfo` creation [here](https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/cfg.rs#L84).

<details>

<summary>Backtrace</summary>
```
thread 'main' panicked at /home/ivor/Documents/Code/rust/bytecodealliance/regalloc2/src/index.rs:155:9:
assertion failed: self.len() > 0
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:145:5
   3: regalloc2::index::InstRange::first
             at /home/ivor/Documents/Code/rust/bytecodealliance/regalloc2/src/index.rs:155:9
   4: regalloc2::cfg::CFGInfo::init
             at /home/ivor/Documents/Code/rust/bytecodealliance/regalloc2/src/cfg.rs:84:60
   5: regalloc2::cfg::CFGInfo::new
             at /home/ivor/Documents/Code/rust/bytecodealliance/regalloc2/src/cfg.rs:46:9
   6: regalloc2::fastalloc::run
             at /home/ivor/Documents/Code/rust/bytecodealliance/regalloc2/src/fastalloc/mod.rs:1290:29
   7: regalloc2::run
             at /home/ivor/Documents/Code/rust/bytecodealliance/regalloc2/src/lib.rs:1614:13
```
</details>


> It's certainly what Cranelift provides in its input, so that will be the path of least resistance for other users as well.

Hmm, it is sure what it feels like. I only really ran into this because I manually hacked up a clif file and ended up with non-consecutive numbers, not sure if anyone else ever ran into this situation? I think it's fine to require them to be consecutive, as long as there's a clear indication that is required and validated for?

cfallin:
Sorry, I'm a bit confused now. You wrote in the initial report

> Running clif-util test from cranelift passes on this file, that probably relabels all ids before passing them to the register allocator?.

But now you're saying

> I only really ran into this because I manually hacked up a clif file and ended up with non-consecutive numbers,

I've confirmed that your CLIF compiles fine with Cranelift, and indeed it does so because Cranelift allocates into new ID spaces for machine-level blocks, instructions and vregs (it's a different IR layer).

I want to make sure I have the precise details of your report right: are you somehow reaching this with CLIF ("because I manually hacked up a clif file")?

iwanders:
> I want to make sure I have the precise details of your report right: are you somehow reaching this with CLIF ("because I manually hacked up a clif file")?

Sorry for the confusion! Running `clif-util` does not result in any problems, the problems merely stem from the id's I pass to regalloc2 through my implementation of the `regalloc2::Function` trait. I do not have these problems when using the `clif-util` binary from the cranelift repository.

When I mentioned 'clif file' I mean anything that ends in `.clif`. In my project I was keeping the ids as they are returned by the [clif_reader](https://docs.rs/cranelift-reader/0.118.0/cranelift_reader/index.html) crate and effectively passing those ids directly to regalloc2. I started with a file from [filetests/isa/x64](https://github.com/bytecodealliance/wasmtime/tree/main/cranelift/filetests/filetests/isa/x64), and modified that in a way that made the indices non consecutive. Then I read that with the reader, and ended up passing the non consecutive ids to regalloc2. Hope this clarifies what I meant, and how I got here.

iwanders:
@cfallin , friendly ping on this, I'm also happy to change this PR to just documentation changes if you'd prefer that.

I forgot to state in my previous message, but the original PR description did include a [minimal](https://gist.github.com/iwanders/433881a4d601e6e7d89fb817d146a0c6) non-working example that shows the original problem.

cfallin:
Sorry, this fell off my radar -- yes, given that a few consecutive-ID assumptions have slipped in, let's document it formally and check it. Happy to review a PR for that (which I guess is close to what you had originally?).

src/lib.rs
```diff
@@ -912,7 +912,7 @@ impl Operand {
     }
 
     /// If this operand is for a fixed non-allocatable register (see
-    /// [`Operand::fixed`]), then returns the physical register that it will
+    /// [`OperandConstraint::FixedReg`]), then returns the physical register that it will
```
cfallin, line 915: This is incorrect (`FixedReg` is not the only aspect of "fixed non-allocatable" operands -- they have a sentinel vreg value too) -- the function reference was stale though. Could you change to `Operand::fixed_nonallocatable`?

cfallin (approved):
Thanks -- almost there, just one comment fix below.

iwanders:
And another small thing, compiling the fuzzer gave a warning in the [pipeline](https://github.com/bytecodealliance/regalloc2/actions/runs/14345887331/job/40215448492?pr=211).

There's now an unused result [here](https://github.com/bytecodealliance/regalloc2/blob/b995300ab4d0071c8551a6839bd7e1a5169cab1e/src/fuzzing/func.rs#L192), I think we can just ignore it with `let _ =`, given that the indices will always be correctly generated and we can't really propagate the `Err` anyway?

src/lib.rs
iwanders: Fixed in 06850ff5b7d3ee40da961155749ee049eb74445f :+1:

cfallin:
Looks good, thanks!

PR #212, opened 2025-04-01
TooManyLiveRegs check: fix `break` that was broken by allocator change.

cfallin:
In #196, various `break`s and early returns` out of a main allocator loop were changed to pass through a point that returned scratch storage to a context for later reuse.

Unfortunately, this refactor introduced a bug: a `break` became a `break 'outer`, and the latter meant that we were skipping some checks.

Fortunately this only happens in a code-path where the given allocation problem has too many constraints, and it means that we get a panic rather than a `TooManyLiveRegs` error cleanly returned, so this is not a security issue. However, it has me a little paranoid and we should carefully re-audit #196's changes when we get a chance.

(Discovered while making changes to ABI code that caused too many defs constrained to regs, resulting in an un-allocatable input; not reachable from mainline Cranelift.)

cc @jakubDoka

### DIFF
--- a/src/ion/process.rs
+++ b/src/ion/process.rs
@@ -1258,8 +1258,7 @@ impl<'a, F: Function> Env<'a, F> {
                                 continue;
                             }
                             if preg_range.from >= range.to {
-                                // Success, return scratch memory to context and finish
-                                break 'outer;
+                                break;
                             }
                             if lr.is_valid() {
                                 if self.minimal_bundle(self.ranges[*lr].bundle) {


fitzgen (approved):
Do you know why the checker didn't catch this during fuzzing?

cfallin:
That's a great question; I suspect it has to do with our fuzzing function generator which adds [up to four](https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/fuzzing/func.rs#L411) operands to any instruction and [up to five](https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/fuzzing/func.rs#L500) clobbers. That would be sufficient to hit this if we had a smaller register environment but we have [32 registers per class](https://github.com/bytecodealliance/regalloc2/blob/925df1b4674435a9322e21912926a68749517861/src/fuzzing/func.rs#L631-L640). All of this provides sufficient coverage for everything but calls; and I suspect we haven't had any hits from higher-level fuzzing because the use-cases we generate were shaped in a particular way (only defs are fixed defs, per the ABI, so everything fits). I'll poke at this when I get some more time.

PR #214, opened 2025-04-06
Redefine a minimal bundle to have only one LiveRange.

cfallin:
When exerting additional pressure on regalloc with bytecodealliance/wasmtime#10502, which can lead to call instructions that have significantly more (`any`-constrained) defs, we hit panics in RA2 where (i) a bundle merged several LiveRanges, (ii) one of these LiveRanges had a fixed-reg constraint on an early use, (iii) this fixed-reg constraint conflicted with a clobber (which always happens at the late-point), (iv) the bundle merged in another LiveRange of some arbitrary def at the late point.

This would make a bundle (which is the atomic unit of allocation) that covers the whole inst, including the late point; and is required to be in the fixed reg; this is unallocatable because the clobber is also at the late point in that reg.

Our allocate-or-split-and-retry logic does not split if a bundle is "minimal". This is meant to give a base case to the retries: when bundles break down into their minimal pieces, any solvable set of constraints should result in allocations.

Unfortunately the "is minimal" predicate definition did not account for multiple LiveRanges, but rather only tested whether the total program-point range of the bundle was over one instruction. If there are multiple LiveRanges, we can still split them off, and the resulting split bundles may cover only half the instruction, avoiding the clobbers.

### DIFF
--- a/src/ion/process.rs
+++ b/src/ion/process.rs
@@ -268,6 +268,7 @@ impl<'a, F: Function> Env<'a, F> {
         let mut fixed = false;
         let mut fixed_def = false;
         let bundledata = &self.ctx.bundles[bundle];
+        let num_ranges = bundledata.ranges.len();
         let first_range = bundledata.ranges[0].index;
         let first_range_data = &self.ctx.ranges[first_range];
 
@@ -277,7 +278,7 @@ impl<'a, F: Function> Env<'a, F> {
             trace!("  -> no vreg; minimal and fixed");
             minimal = true;
             fixed = true;
-        } else {
+        } else if num_ranges == 1 {
             for u in &first_range_data.uses {
                 trace!("  -> use: {:?}", u);
                 if let OperandConstraint::FixedReg(_) = u.operand.constraint() {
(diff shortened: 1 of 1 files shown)


cfallin:
This fixes one of the fuzzbugs that came in (https://issues.oss-fuzz.com/issues/408523840) -- likely the others too, but I haven't checked yet. I'm fuzzing locally with `ion_checker` and will leave it running on 32 threads over the weekend to validate this.

cfallin:
I fuzzed this (with the `ion_checker` target, `-s none` for more speed, 16c/32t on a Ryzen 9950x system) for 25h36m with no checker failures found; ran up to:

```
#540595696: cov: 27746 ft: 26948 corp: 3367 exec/s 117 oom/timeout/crash: 0/50/0 time: 92203s job: 9713 dft_time: 0
```

alexcrichton:
Would it make sense to beef up the fuzzer in this repo to catch this as well? (assuming it isn't possible for the preexisting fuzzer to eventually catch this)

cfallin:
Yes definitely, I need to look into that -- it won't be super-simple though, as there isn't really a separate oracle for "is this set of constraints allocatable", so I need to do some thinking about how to generate more interesting sets of constraints by construction that work. In other words I don't want to block merging this fix on that as it will be a bit of a deeper task.

fitzgen (approved):
+1 on trying to cover these cases in the fuzz target when possible

PR #216, opened 2025-04-07
Fuzzing: add "callsite-ish constraints".

cfallin:
This is a followup from #214. We have use-cases in Cranelift now where we generate constraints on call instructions that have:

- A number of fixed-reg uses at "early";
- A number of fixed-reg defs at "late";
- A number of register clobbers, potentially overlapping with uses;
- A number of "any"-constrained defs at "late", to add register pressure.

The existing fuzzing function generator would create at most 3 operands, and would add *either* fixed-reg constraints *or* clobbers, not both.

This PR adds a "callsite-ish constraints" branch as a subcase of the fixed-reg choice that also adds the "any" defs to add register pressure, and adds new clobbers that don't interfere with existing constraints. It also loosens the logic around constructing fixed-reg constraints: if a conflict is found with some arbitrarily-chosen constraint, it keeps going to the next operand and tries to make it fixed, rather than giving up on further operands.

### DIFF
--- a/fuzz/fuzz_targets/fastalloc_checker.rs
+++ b/fuzz/fuzz_targets/fastalloc_checker.rs
@@ -25,6 +25,7 @@ impl Arbitrary<'_> for TestCase {
                     fixed_nonallocatable: true,
                     clobbers: true,
                     reftypes: false,
+                    callsite_ish_constraints: true,
                 },
             )?,
         })
--- a/fuzz/fuzz_targets/ion_checker.rs
+++ b/fuzz/fuzz_targets/ion_checker.rs
@@ -25,6 +25,7 @@ impl Arbitrary<'_> for TestCase {
                     fixed_nonallocatable: true,
                     clobbers: true,
                     reftypes: true,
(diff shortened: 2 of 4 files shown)


cfallin:
I'm fuzzing locally on top of a branch without #214 to see if this will find the bug. That bug was in the middle of an extremely large function body, so I'm not sure how long the search might take, but in theory the constraints on generated functions should be capable of covering the case that we hit there.

fitzgen (approved):
Nice! Looking forward to seeing if this can independently find that bug

cfallin:
This has at least fuzzed for a few CPU-hours without finding any *other* issues, so I'll go ahead and merge. I can verify (from playing with the constraint generation and making it too loose) that the new case is being explored; it seems just not to have gotten the right combo of register pressure and liveranges to get a bundle-merge at the right point yet.

(Sometime in the further future it would be neat to hook the bundle-merging and other heuristic bits into a chaos control plane of sorts, like we have in Cranelift, to more easily explore edge cases like this)

Issue #219, opened 2025-04-09
Performance regression / infinite loop with 0.11.3 release


Mrmaxmeier:
Hi,

I'm using Cranelift with `opt_level=speed_and_size` in a project and have run into a hang after updating to the current main branch of the wasmtime/cranelift repository. Git bisect points to this commit: [b772c31ef](https://github.com/bytecodealliance/wasmtime/commit/b772c31efee95ef409e4c4ec9ea77bb857b5bd7b) "Cranelift: update to regalloc2 0.11.3."

My original reproducer is a single large Cranelift function (~30kb) that compiled almost instantly with the previous commit but hangs after the regalloc2 update. I'm not sure if the regression can also be triggered by WebAssembly modules.

```console
$ time ./clif-util-bad compile --target=x86_64 --set opt_level=speed regression.clif
^C
Executed in   41.20 secs
$ time ./clif-util-good compile --target=x86_64 --set opt_level=speed regression.clif
Executed in   16.72 millis
```

Thanks!

CC #214 since this is the only change between 0.11.2 and 0.11.3

[regression.clif.txt](https://github.com/user-attachments/files/19668867/regression.clif.txt)

[reduced.clif.txt](https://github.com/user-attachments/files/19670982/reduced.clif.txt)

BTW: Is there a way to reduce hangs/timeouts with `clif-util bugpoint`? I've hacked something together using `shrinkray` but I'm not fully convinced of the process. Cranelift-native tooling should probably be more effective 🙂

Mrmaxmeier:
Ah whoops, should've checked between writing the issue and running the test case minimizer 🙂 
The test file might still be useful though.
Closing as a duplicate of #219

cfallin:
No worries and the additional test case is useful, thanks! I'm actively looking into this now.

PR #220, opened 2025-04-11
Fix minimal-bundle splitting to avoid infinite loops.

### LINKED ISSUE #218: Release v0.11.3 (but not v0.11.2) hangs while compiling a few specific wasm modules
Using the latest wasmtime (v31), compiling the attached wasm module with `wasmtime compile` hangs after updating regalloc to v0.11.3. It compiles instantly when using regalloc v0.11.2 or below.

Specifically:

1. Clone wasmtime.
2. Checkout the `release-31.0.0` branch.
3. Run `cargo update regalloc2` to update to v0.11.3.
4. Run `cargo run --release -- compile example.wasm` (`--release` isn't necessary but it demonstrates that it's not just slow due to running in debug mode).

See [example.zip](https://github.com/user-attachments/files/19670032/example.zip) (extract it first, GitHub doesn't allow uploading raw wasm files).

I've also tried patching in the latest main branch of `regalloc2` and that doesn't seem to help. This points to a bug in https://github.com/bytecodealliance/regalloc2/pull/214 that hasn't been fixed yet.

cfallin:
This PR reworks minimal-bundle logic to be consistent between property computation and splitting, and overall simpler.

To recap: a "minimal bundle" is an allocation bundle that is as small as possible. Because RA2 does backtracking, i.e., can kick a bundle out of an allocation to make way for another, and because it can split, i.e. create more work for itself, we need to ensure the algorithm "bottoms out" somewhere to ensure termination. The way we do this is by defining a "minimal bundle": this is a bundle that cannot be split further. A minimal bundle is never evicted, and any allocatable input (set of uses/defs with constraints), when split into minimal bundles, should result in no conflicts. We thus want to define minimal bundles as *minimally* as possible so that we have the maximal solving capacity.

For a long time, we defined minimal bundles as those spanning a single instruction (before- and after- progpoints) -- because we cannot insert moves in the middle of an instruction. In #214, we updated minimal-bundle splitting to avoid putting two different unrelated uses in a single-instruction bundle, beacuse doing so and calling it "minimal" (unsplittable) can artificially extend the liverange of a use with a fixed-reg constraint at the before-point into the after-point, causing an unsolveable conflict. This was triggered by new and tighter constraints on callsites in Cranelift after
bytecodealliance/wasmtime#10502 (merging retval defs into calls) landed.

Unfortunately this also resulted in an infinite allocation loop, because the definition of "minimal bundle" did not agree between the split-into-minimal-bundles fallback/last-ditch code, and the bundle property computation. The splitter was splitting as far as it was willing to go, but our predicate didn't consider those bundles minimal, so we continued to re-attempt splitting indefinitely.

While investigating this, I found that the minimal-bundle concept had accumulated significant cruft ("the detritus of dead fuzzbugs") and this tech-debt was making things more confusing than not -- so I started by clearly defining what a minimal bundle *is*. Precisely:

- A single use, within a single LiveRange;
- With that LiveRange having a program-point span consistent with the use:
  - Early def: whole instruction (must live past Late point so it can reach its uses; moves not possible within inst);
  - Late def: Late point only;
  - Early use: Early point only;
  - Late use: whole instruction (must be live starting at Early so the value can reach this use; moves not possible within inst).

This is made easier and simpler than what we have before largely because the minimal-bundle splitter aggressively puts spans of LiveRange without uses into the spill bundle, and because we support overlapping LiveRanges for a vreg now (thanks Trevor!), so we can rely on having *some* connectivity between the def and its uses even if we aggressively trim LiveRanges in the minimal bundles down to just their defs/uses.

The split-at-program-point splitter (i.e., not the fallback split-into-minimal-bundles splitter) also got a small fix related to this: it has a mode that was intended to "split off one use" if we enter with a split-point at the start of the bundle, but this was really splitting off all uses at the program point (if there are multiple of the same vreg at the same program point). In the case that we still need to split these apart, this just falls back to the minimal-bundle splitter now.

Fixes #218, #219.

### DIFF
--- a/src/ion/mod.rs
+++ b/src/ion/mod.rs
@@ -97,6 +97,9 @@ impl<'a, F: Function> Env<'a, F> {
         self.process_bundles()?;
         self.try_allocating_regs_for_spilled_bundles();
         self.allocate_spillslots();
+        if trace_enabled!() {
+            self.dump_state();
+        }
         let moves = self.apply_allocations_and_insert_moves();
         Ok(self.resolve_inserted_moves(moves))
     }
--- a/src/ion/process.rs
+++ b/src/ion/process.rs
@@ -19,7 +19,7 @@ use super::{
 };
 use crate::{
     ion::data_structures::{
-        CodeRange, BUNDLE_MAX_NORMAL_SPILL_WEIGHT, MAX_SPLITS_PER_SPILLSET,
+        CodeRange, Use, BUNDLE_MAX_NORMAL_SPILL_WEIGHT, MAX_SPLITS_PER_SPILLSET,
         MINIMAL_BUNDLE_SPILL_WEIGHT, MINIMAL_FIXED_BUNDLE_SPILL_WEIGHT,
     },
     Allocation, Function, Inst, InstPosition, OperandConstraint, OperandKind, PReg, ProgPoint,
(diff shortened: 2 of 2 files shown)


cfallin:
I'm fuzzing this locally; hit a few fuzzbugs immediately with my initial fix (which led to some of the other cases included here), but now ~15 min in on 32 threads, I'll leave it running at least for the rest of today.

Amanieu:
> A single use, within a single LiveRange;

What about the case where the same vreg is used multiple times in the same instruction (as a "use", not a "def") *without* a fixed-reg constraint? That would result in a minimal bundle with multiple uses.

cfallin:
@Amanieu yes, we could start to carve out exceptions like that, but this is meant to be a simplest possible definition, and the way that we got into this corner was by accreting complexity over the years, so I'd prefer to keep this very simple if possible! Ideally minimal-bundle splitting doesn't kick in very often anyway.

Amanieu:
The problem is that your proposed definition can result in bundles that are not minimal and cannot be split any further.

I would argue that a better definition would be:
- A single LiveRange (but any number of uses)
- Does not cross an instruction boundary (starts & ends on the same instruction).

cfallin:
I don't think that's actually true: a bundle with multiple uses can always be split by `split_into_minimal_bundles()`; it creates a new LiveRange and bundle for every use. Please see how I've reworked it in this PR: it literally loops through uses and creates a new LR/bundle for each one.

(I think you're still thinking of splitting as defined by a split-point; splitting here is literally scattering uses to new LRs.)

fitzgen (approved):
LGTM, thanks!

cfallin:
Fuzzer still going and currently at

```
#69835175: cov: 30601 ft: 27476 corp: 3111 exec/s 160 oom/timeout/crash: 0/0/0 time: 13625s job: 1578 dft_time: 0
```

or about 3.8 hours at 32 threads; so I'm going to call this good to merge (and then we'll get more fuzzing in Cranelift!).

Issue #222, opened 2025-04-13
Algorithm::Ion may panic with message "Could not allocate minimal bundle, but the allocation problem should be possible to solve"


MaxVerevkin:
`regalloc2 v0.12.0`

```text
thread 'main' panicked at /home/maxv/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.12.0/src/ion/process.rs:1253:17:
Could not allocate minimal bundle, but the allocation problem should be possible to solve
```

Function which causes a panic:

<details>
<summary> Debug repr </summary>

  ```text
  {
    machine_env: MachineEnv {
      preferred_regs_by_class: [
          [
              PReg(hw = 0, class = Int, index = 0),
              PReg(hw = 1, class = Int, index = 1),
              PReg(hw = 2, class = Int, index = 2),
          ],
          [],
          [],
      ],
      non_preferred_regs_by_class: [
          [],
          [],
          [],
      ],
      scratch_by_class: [
          None,
          None,
          None,
      ],
      fixed_stack_slots: [],
  }
    spillslot_size(Int): 1
    spillslot_size(Float): 1
    spillslot_size(Vector): 1
    multi_spillslot_named_by_last_slot: false
    allow_multiple_vreg_defs: false
    block0(): # succs:[] preds:[]
      inst0: op Def: v0i any
      inst1: op Def: v1i fixed(p0i), Use: v0i fixed(p1i), Clobber: p0i, Clobber: p1i, Clobber: p2i
      inst2: ret Use: v1i fixed(p0i)
  }
  ```
</details>

<details>
<summary> JSON repr </summary>

```json
{
  "machine_env": {
    "preferred_regs_by_class": [
      [
        {
          "bits": 0
        },
        {
          "bits": 1
        },
        {
          "bits": 2
        }
      ],
      [],
      []
    ],
    "non_preferred_regs_by_class": [
      [],
      [],
      []
    ],
    "scratch_by_class": [
      null,
      null,
      null
    ],
    "fixed_stack_slots": []
  },
  "entry_block": 0,
  "insts": [
    {
      "op": "Op",
      "operands": [
        {
          "bits": 8388609
        }
      ],
      "clobbers": {
        "bits": [
          0,
          0,
          0,
          0
        ]
      }
    },
    {
      "op": "Op",
      "operands": [
        {
          "bits": 2155872256
        },
        {
          "bits": 2197815297
        }
      ],
      "clobbers": {
[...]

PR #223, opened 2025-04-13
API documentation update: add a note about clobber/def conflicts.

cfallin:
This arose in #222. It has been the case for a while that we do not permit clobbers and defs to name the same physical register; this is because clobbers become defs internally. Unfortunately the doc-comment on the relevant part of the `Function` trait was out-of-date.

Ideally we would also detect this during validation or liverange construction rather than panic'ing later, but that is a separate concern (panic-cleanliness).

Note that this doesn't affect Cranelift; this is a concern mainly for docs correctness and for third-party users.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1193,12 +1193,11 @@ pub trait Function {
     /// used as a vreg output, or fixed physical registers used as
     /// temps within an instruction out of necessity.
     ///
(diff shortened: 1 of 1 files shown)


cfallin:
cc @MaxVerevkin

fitzgen (approved):

Issue #217, opened 2025-04-08
fastalloc does not support arbitrary numbers of `any` defs allocating into spillslots


cfallin:
After bytecodealliance/wasmtime#10502, a necessary prerequisite for exception handling in Cranelift (and thus Wasmtime), we generate all return-value defs for callsites as defs on the call instruction, with an `any` constraint. The default (backtracking) allocator is able to allocate defs past the number of registers into spillslots directly. However, it appears that fastalloc cannot do this: it aborts if a def with an `any` constraint cannot allocate a register.

See [this fuzzbug](https://oss-fuzz.com/testcase-detail/5433312476987392) for a reproduction.

cfallin:
I'm filing this as a tracking issue and will disable fastalloc in Cranelift for now; it's a tier-3 feature and I don't have the cycles to update the allocator unfortunately. At some later time we may be able to fix this limitation and then re-enable it.

d-sonuga:
Hey @cfallin, is this the same bug?

```
thread '<unnamed>' panicked at /Users/d-sonuga/repos/regalloc2/src/fastalloc/mod.rs:154:56:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==86022== ERROR: libFuzzer: deadly signal
    #0 0x00010723dcc0 in __sanitizer_print_stack_trace+0x28 (librustc-nightly_rt.asan.dylib:arm64+0x5dcc0)
    #1 0x000105a9eef0 in fuzzer::PrintStackTrace()+0x30 (fastalloc_checker:arm64+0x101006ef0)
    #2 0x000105a92154 in fuzzer::Fuzzer::CrashCallback()+0x54 (fastalloc_checker:arm64+0x100ffa154)
    #3 0x0001837b8620 in _sigtramp+0x34 (libsystem_platform.dylib:arm64+0x3620)
    #4 0x00018377e888 in pthread_kill+0x124 (libsystem_pthread.dylib:arm64+0x6888)
    #5 0x000183687c5c in abort+0x78 (libsystem_c.dylib:arm64+0x78c5c)
    #6 0x000105be42c8 in std::sys::pal::unix::abort_internal::hc6a174fee9c4fd65+0x8 (fastalloc_checker:arm64+0x10114c2c8)
    #7 0x000105be34ac in std::process::abort::he28388c695e4448f+0x8 (fastalloc_checker:arm64+0x10114b4ac)
[...]

PR #229, opened 2025-06-21
Fix fastalloc edge case with nonallocated scratch register

squaaawk:
Resolves the existing fuzzbugs I've found running the fuzzer. I've also de-duplicated some code related to scratch register allocation, which is a nice simplification.

I don't have access to the fuzzbug in #217, so I can't check if this PR resolves that issue as well.


cfallin (approved):
Looks reasonable -- thanks!

PR #233, opened 2025-08-24
Arbitrary Any Defs and Stack-Only Constraints for Fastalloc

d-sonuga:
Fixes https://github.com/bytecodealliance/regalloc2/issues/217.

- Adds support for an arbitrary number of Any def operands
- Adds support for stack-only operands

Benchmarking this branch against the implementation on the main branch with sightglass, most benchmarks have no significant difference while a few others have a speed difference of about 1-1.83x, with the main branch being faster on some and this branch being faster on others.


cfallin (approved):
This looks reasonable -- thanks very much for making this update!

PR #235, opened 2025-09-17
Refactor `RegTraversalIter`

abrown:
This is a non-functional change to refactor `RegTraversalIter` to make it more amenable for future commits adding limits to which registers can be allocated. This change does change the API of `RegTraversalIter`, though.

It boils down to several parts:
- using `Cursor` internally: much of the cursor-advancing logic is duplicated between the preferred and non-preferred register list. This change adds a new, private struct--`Cursor`--to handle this. This refactoring also reduces the scope of the borrow we take down to only what is necessary (i.e., the actual register lists).
- changing parameter order: this also passes through the arguments constructing a `RegTraversalIter` in the order they will be checked during iteration: first it emits the fixed register, then the hinted register, then the offset-based lookup. This is a clarity tweak.
- only allow one hint register: `RegTraversalIter` allowed for two hint registers but the library never uses the second one. This change switches to only using a single register, with no other change in behavior.
- minor simplifications: e.g., we can use `Option::take` to avoid a more verbose access (and removal) of the fixed register.

Another possible change here would be to pass the hinted register as an `Option<PReg>` with `None` to indicate no hint. This seems more conventional and a quick audit of the code supports this but I avoided this out of an abundance of caution.


cfallin (approved):
This is much cleaner -- thanks very much!

I'm not sure why the original implementation didn't use `Option<PReg>` for the `hint` field rather than the separate flag and I agree that'd be cleaner -- feel free to make that change as well.

cfallin:
(I'll merge this now but happy to take said additional cleanup as part of another PR.)

PR #239, opened 2025-09-18
Add limits to allocatable registers

abrown:
This change adds the ability to set an upper limit on the hardware encodings of operands to an instruction. This will be publicly available via a new constraint: `OperandConstraint::Limit`.

Why is this necessary? In x86, different instructions can have different sets of available registers; e.g., consider `ADDPS`, which can address 16 registers using its legacy SSE encoding and `VADDPS`, which can address 32 registers in its AVX512 EVEX encoding. With the introduction of [APX](https://www.intel.com/content/www/us/en/developer/articles/technical/advanced-performance-extensions-apx.html), this difference will propagate to many scalar instructions (i.e., addressing 16 registers normally but 32 registers with APX EVEX encodings). The ability to set an upper limit allows us to configure, per instruction, which set of registers an instruction can allocate to.

This first commit handles the necessary merging, splitting, and register iteration in `ion` using this new constraint, but does not implement every piece yet:
- for now, we disable the use of `OperandConstraint::Limit` in `fastalloc`; more discussion is needed to reconcile recent changes, which have some conceptual overlap (e.g., `available_pregs_for_regs`)
- this does not yet generate `Limit` operands during fuzzing; I have changes queued up to do this separately, since I suspect that will need some fine-tuning of the fuzz parameters and I would like it to be revertible separately (i.e., the addition of the `Limit` dimension makes generating a valid, allocatable function a bit more n-queens-like)

Each commit has more details.


src/ion/data_structures.rs
```diff
@@ -203,6 +203,7 @@ pub struct LiveBundle {
     pub allocation: Allocation,
     pub prio: u32, // recomputed after every bulk update
     pub spill_weight_and_props: u32,
+    pub limit: Option<usize>,
```
cfallin, line 206: The size of these structs is pretty important for performance (cache residency etc) -- instead of an `Option<usize>` could we use an `Option<u8>` here? (The number of physical registers is already constrained to 128 so this should always fit?)

cfallin (approved):
This looks great overall -- thanks very much for working on this!

I have a few random thoughts below, and also a high-level question: have you checked compile-time performance with these changes in Cranelift/Wasmtime (but with no limits actually used)? I ask because this does add a little data to the bundle info and one more scan over uses; hopefully it's minimal.

abrown:
I ran some benchmarks, specifically the [`default.suite`] in Sightglass. First I built Wasmtime using the commit updating `regalloc2` to v0.13.2 (https://github.com/bytecodealliance/wasmtime/commit/9362e47987363c350a8d9d09aa56677425496fb7) into `before/libengine.so`. Then I built a Wasmtime using this PR in a special branch, [`bench-regalloc-limits`], into `after/libengine.so`. Measuring on a single core, I do not see any difference in performance:

[`default.suite`]: https://github.com/bytecodealliance/sightglass/blob/main/benchmarks/default.suite
[`bench-regalloc-limits`]: https://github.com/abrown/wasmtime/tree/bench-regalloc-limits

```console
$ taskset --cpu-list 12 cargo run -- benchmark \
        --engine /.../before/libengine.so \
        --engine /.../after/libengine.so \
        --benchmark-phase compilation \
        /.../benchmarks/default.suite
...
compilation :: cycles :: /.../benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [194804612 200477001.54 209554108] after/libengine.so
  [195231061 201591892.87 212052246] before/libengine.so

compilation :: cycles :: /.../benchmarks/pulldown-cmark/benchmark.wasm

  No difference in performance.

  [421197011 432378860.95 496054031] after/libengine.so
  [421450247 430139316.32 444803028] before/libengine.so

compilation :: cycles :: /.../benchmarks/spidermonkey/benchmark.wasm

  No difference in performance.

  [14598855627 14742970537.75 14959672535] after/libengine.so
  [14602280462 14792858580.32 15978671047] before/libengine.so
```

This is measured before any of the changes suggested above (e.g., `Option<usize>` to `Option<u8>`, `Cursor` constraints).

cfallin:
Excellent -- thanks for double-checking!

PR #240, opened 2025-09-18
Fastalloc Def branch arg on branch & Bug Fixes

d-sonuga:
Resolves https://github.com/bytecodealliance/wasmtime/issues/11545 and https://github.com/bytecodealliance/wasmtime/issues/11544.

- Add support for any, fixed-reg and stack-only branch arguments being defined in its branch instruction.
- Remove over-constraint on reservation of registers for operands with any-reg constraints. Instead, use counters to decide when it is safe to allocate registers to operands with no constraints.
- Correct `allocd_within_constraint` to account for allocation of clobbers to late phase registers.
- Correct `select_suitable_reg_in_lru` to only allocate available pregs in both late and early phases to early defs and late uses.
- Allocate late operands first, followed by early operands, instead of defs then uses.
- Remove over-constraint on available registers by removing clobbers only from the late available register set.

Previously, operands with any-reg constraints all got their registers reserved as if they were all late uses so as to avoid a situation where a register meant for an operand valid for both early and late phases is allocated to an operand valid only in an early phase or a late phase, but not both, potentially leaving no valid registers for an early & late phase operand.
This is an over-constraint that led to this issue https://github.com/bytecodealliance/wasmtime/issues/11544.
This is resolved by completely ditching the reservation of any-reg operands in favor of using counters to determine whether or not it is safe to allocate registers to operands with no constraints.

Another issue:
```
use v0 fixed(p0), def v1 fixed(p0), use late v0 any
```
In this scenario, p0 is fixed to both v1 and v0, but that shouldn't be a problem because they are in different phases. Prior to this PR, this was problematic because all defs were allocated first, then uses resulting in an allocation order in the above example that looked like this:
```
p0 -> v1 (this is a def, so it's freed and vreg_allocs[v1] is set to none)
p0 -> v0 (vreg_allocs[v0] = p0)
p0 -> v0 (vreg_allocs[v0] is p0, which is within constraints, so it is selected)
```
Which is incorrect. The root cause is that during allocation, vreg_allocs[vi] tells the current allocation of some register vi - but when the late v0 operand is being allocated, vreg_allocs[vi] tells the allocation of v0 in the early phase of the instruction, not the late phase, and since allocation proceeds in reverse, this is an incorrect order. It should always proceed from the late phase to the early phase. To resolve this, instead of all def operands being allocated first, then use operands, it's the late operands that are allocated first, followed by the early operands. This is still safe because the reason def operands were allocated first is because registers allocated to late def operands can be reused by early use operands, and in this processing order, this order will still remain this same.

Fuzzed overnight for 8-9 hours.
I also ran Wasmtime's tests. Most pass. The ones that didn't pass didn't seem to fail because of register allocation - for example, the disas test checks against hardcoded output.

### DIFF
--- a/src/fastalloc/mod.rs
+++ b/src/fastalloc/mod.rs
@@ -491,12 +491,12 @@ impl<'a, F: Function> Env<'a, F> {
             && self.edits.scratch_regs[class].is_none()
         {
             self.alloc_scratch_reg(inst, class, pos)?;
-            self.num_available_pregs[ExclusiveOperandPos::Both][class] =
-                0i16.max(self.num_available_pregs[ExclusiveOperandPos::Both][class] - 1);
-            self.num_available_pregs[ExclusiveOperandPos::LateOnly][class] =
-                0i16.max(self.num_available_pregs[ExclusiveOperandPos::LateOnly][class] - 1);
-            self.num_available_pregs[ExclusiveOperandPos::EarlyOnly][class] -=
-                0i16.max(self.num_available_pregs[ExclusiveOperandPos::EarlyOnly][class] - 1);
+            let dec_clamp_zero = |x: &mut i16| {
+                *x = 0i16.max(*x - 1);
+            };
+            dec_clamp_zero(&mut self.num_available_pregs[ExclusiveOperandPos::Both][class]);
+            dec_clamp_zero(&mut self.num_available_pregs[ExclusiveOperandPos::EarlyOnly][class]);
+            dec_clamp_zero(&mut self.num_available_pregs[ExclusiveOperandPos::LateOnly][class]);
         }
         self.edits.add_move(inst, from, to, class, pos);
         Ok(())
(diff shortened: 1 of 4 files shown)


src/fastalloc/mod.rs
```diff
@@ -406,6 +491,12 @@ impl<'a, F: Function> Env<'a, F> {
             && self.edits.scratch_regs[class].is_none()
         {
             self.alloc_scratch_reg(inst, class, pos)?;
+            self.num_available_pregs[ExclusiveOperandPos::Both][class] =
+                0i16.max(self.num_available_pregs[ExclusiveOperandPos::Both][class] - 1);
```
cfallin, line 495: We can use `saturating_sub()` here and below, I think, to avoid the underflow on `-`.

src/fastalloc/mod.rs
```diff
@@ -1260,6 +1419,67 @@ impl<'a, F: Function> Env<'a, F> {
...
+                        first_inst,
+                        from,
+                        to,
+                        block_param.class(),
+                        InstPosition::Before,
+                    )?;
+                }
+                break;
+            }
+        }
```
cfallin, line 1482: Can we check somewhere above that there is only one predecessor in the case that a branch's def forces a move insertion here? That should always be the case when such defs are on conditional branches (because we require split critical edges) and is thus sufficient for Cranelift's `try_call` at least; and it avoids the incorrectness that would otherwise result.

cfallin (approved):
Thanks very much for these fixes!

The "move in successor if predecessor def'd value in branch at fixed reg" exception to the always-in-spillslots-at-block-start invariant is kind of unfortunate, but I think I agree that it's the most pragmatic solution here. The only other way out seems to be to generalize the inter-block state, so we actually track locations of every value, which gives up a lot of the efficiency advantages that SSRA has.

src/fastalloc/mod.rs
d-sonuga: They are `i16`s and`saturating_sub` is relative to the type's minimum value, but I've clarified that code anyways.

cfallin: Ah, that's right; thanks for the small refactor there!

PR #244, opened 2025-09-22
Run fuzzing during `cargo test`; reorganize fuzzers

abrown:
This change moves the meat of the code for fuzzing `regalloc2` into the main crate's `fuzzing` module. This has several minor benefits (easier to find, better LSP detection, same workspace for `cargo fmt|clippy`) but the major one is we can not only build, but also _run_ some fuzz tests with:

```
cargo test --features fuzzing
```

The fuzz modules now use `arbtest` to run one second worth of fuzzing. This should be helpful both during development and in CI.

The second part of this change is a reorganization of the fuzz targets themselves. In looking at them carefully, I realized:
- `ion` and `ion_checker` were doing essentially the same thing; in fact, we probably always want to run the checker after allocation
- `ssagen` was not necessary: we could get the same functionality by flipping the `enable_ssa_checker` flag on `fastalloc|ion::run()`
- we never fuzzed with annotations off (not sure why we would not want to but... shrug)

This led to randomly flipping the annotations and SSA checking flags, renaming `ion_checker.rs -> ion.rs` (replacing `ion.rs`) and `fastalloc_checker.rs -> fastalloc.rs`. `ssagen.rs` is now gone, subsumed by now being able to check the SSA in both `ion.rs` and `fastalloc.rs`.

This naming change will undoubtedly require some re-configuration in OSSFuzz but that can happen later.


abrown:
@cfallin, I went back and forth on removing the `ssagen` target: my sense was that what it does is already being covered by turning on the SSA checker for `ion` and `fastalloc`, but I can also understand it being useful separately, especially during development. If you want me to keep that one I'll undo that part of this PR.

The other incomplete part with this PR is the documentation in `doc/ION.md`: if you agree with the changes here, then I can go and simplify some of that documentation.

PR #245, opened 2025-09-22
Avoid extra block calculation during fuzz generation

abrown:
In the process of building up a control-flow graph of blocks within a generated function, these lines appear to re-calculate the edge links between successors and predecessors. This _should_ be taken care of by `FuncBuilder::add_edge` or so it would appear. This change removes the extra calculation, assuming it was a leftover from some incomplete refactoring during the initial work on this crate.


abrown:
...and I just stopped it at:

```
#25457754: cov: 4597 ft: 27051 corp: 4095 exec/s: 260 oom/timeout/crash: 0/0/0 time: 5687s job: 442 dft_time: 0
```

Everything seems fine if in fact `ion_checker` is the right target to run.

PR #246, opened 2025-11-09
Handle case where livein is defined on a branch instruction

d-sonuga:
Fixes https://github.com/bytecodealliance/wasmtime/issues/11850

This PR does two things:
- Rearranges some code to reduce repetition 
- Handles the case where a livein vreg is defined on a branch instruction in fastalloc


cfallin (approved):
Looks like a reasonable point-fix -- thanks!

PR #260, opened 2026-04-15
Release 0.15.1.

cfallin:
Includes #259 (version-bump of deps).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = ["regalloc2-tool"]
 
 [package]
 name = "regalloc2"
-version = "0.15.0"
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
To confirm, hashbrown maps don't show up in the public API anywhere right? If so this'd need to be a major version bump, but if not a minor works

cfallin:
Yes, that's right -- the public API is only the `Function` trait and various enums/structs that have at most slices or `Vec`s.

cfallin:
[Released on crates.io](https://crates.io/crates/regalloc2/0.15.1)

Issue #87, opened 2022-09-23
regalloc2 makes too many memory allocations


Amanieu:
I used [heaptrack](https://github.com/KDE/heaptrack) to analyze the memory allocation patterns of my compiler which uses regalloc2. I used a large benchmark (LLVM library) which has 1257955 basic blocks in 104913 functions. Out of a total of 46780667 allocations, 46779011 of them come from regalloc2 and only 1656 come from other parts of the compiler.

My compiler uses a similar structure [^1] to Cranelift where a [`Context`](https://docs.rs/cranelift-codegen/latest/cranelift_codegen/struct.Context.html) structure allows memory allocations to be reused across compilation units. Specifically, this allows various `Vec`s and `HashMap`s to grow to a sufficient size only once while being reused many times for compiling multiple functions.

[^1]: In fact I am using the `cranelift-entity` crate directly, it's very well written.

regalloc2 unfortunately doesn't follow this design:
- A fresh `Env` structure is allocated and freed for each function. No memory allocations are preserved.
- Heavy use of temporary `SmallVec`s and `HashSet`s. Heaptrack reports that 19865290 allocations (42% of the total) are "temporary". These are allocations which are immediately followed by a free, with no other allocations in between.
- Heavy use of `BTreeMap` which needs to allocate many nodes separately and cannot reuse memory unlike `HashMap` and `Vec`.

## Heaptrack results

Here are the major allocation hotspots shown by heaptrack:

In `try_to_allocate_bundle_to_reg`:
- 16158324 temporary allocations from the `FxHashSet`.
- 9677435 allocations from inserting liveranges into `PRegData::allocations` (`LiveRangeSet` backed by a `BTreeMap`).

In `merge_bundles`:
- 4478881 allocations from pushing a `VReg` to `SpillSet::vregs` (`SmallVec`).
- 1895443 allocations from appending `LiveRangeListEntry`s to `LiveBundle::ranges` (`SmallVec`).

In `insert_use_into_liverange`:
- 2578957 allocations from pushing a `Use` to `LiveRange::uses` (`SmallVec`).

In `add_liverange_to_vreg`:
- 1432853 allocations from pushing a `LiveRangeListEntry` to `VRegData::ranges` (`SmallVec`).

In `create_pregs_and_vregs`:
- 948582 allocations from pushing to `Env::vregs` and `Env::allocs` and `Env::inst_alloc_offsets`. (`Vec`)

In `Env::new`:
[...]

PR #261, opened 2026-06-05
perf: Replace HashSet with generational vec

michael-weigelt:
When compiling certain types of functions (e.g. with many locals), a lot of time is spent in `try_to_allocate_bundle_to_reg`, in particular on `HashSet::insert`. This data structure is only used for deduplicating and the cost can be quite high (hashing, allocations, bad cache properties).

Replacing it with a generational vector gives significant improvements for my ad-hoc benchmark (deliberately compiled with no optimizations, because my use-case currently demands it): 
Compiling a Wasm function with X locals before and after this patch: 
```
1000 locals: 780ms compilation -> 363ms   -53%
10k  locals:  66s compilation ->  23s     -65%
40k  locals: 952s compilation ->  356s    -62%
```
The Sightglass benchmarks are mixed, see the comment below.

### DIFF
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -447,32 +447,27 @@ impl core::ops::IndexMut<VReg> for VRegs {
 /// A dedup set of `LiveBundleIndex` values that avoids hashing.
 ///
 /// Each bundle index is mapped to a slot in a dense array holding the
-/// generation at which it was last inserted. `clear` simply bumps the
-/// current generation (O(1) in the common case), and `insert` is a
-/// single bounds-checked compare-and-store. This is a drop-in
-/// replacement for the previous `FxHashSet<LiveBundleIndex>` that is
-/// much cheaper when a bundle conflicts with many others (e.g. a
-/// function with many locals).
+/// generation at which it was last inserted. `clear` bumps the current
+/// generation and `insert` updates the stamp for the given bundle and
+/// resizes the array if necessary.
+/// This is a drop-in replacement for the previous `FxHashSet<LiveBundleIndex>`
+/// that is much cheaper when a bundle conflicts with many others (e.g. for
+/// functions with many locals).
 #[derive(Clone, Debug, Default)]
 pub struct ConflictSet {
     // Generation at which each bundle was last inserted. The value `0`
     // means "never inserted"; `generation` is therefore always >= 1
     // after the first `clear`.
-    stamps: Vec<u32>,
-    generation: u32,
+    stamps: Vec<u64>,
+    generation: u64,
 }
 
 impl ConflictSet {
-    /// Empty the set. O(1) except on the rare generation wraparound.
+    /// Empty the set by bumping the generation.
+    /// Every value below the new generation is now stale.
     #[inline]
     pub fn clear(&mut self) {
-        self.generation = self.generation.wrapping_add(1);
-        if self.generation == 0 {
-            // Wrapped around; reset stamps so stale entries don't read
-            // as present, and skip the reserved `0` generation.
-            self.stamps.iter_mut().for_each(|s| *s = 0);
-            self.generation = 1;
-        }
+        self.generation += 1;
     }
 
     /// Insert `bundle`. Returns `true` if it was not already present.
--- a/src/ion/data_structures.rs
+++ b/src/ion/data_structures.rs
@@ -444,6 +444,53 @@ impl core::ops::IndexMut<VReg> for VRegs {
     }
 }
 
+/// A dedup set of `LiveBundleIndex` values that avoids hashing.
+///
+/// Each bundle index is mapped to a slot in a dense array holding the
+/// generation at which it was last inserted. `clear` simply bumps the
+/// current generation (O(1) in the common case), and `insert` is a
+/// single bounds-checked compare-and-store. This is a drop-in
+/// replacement for the previous `FxHashSet<LiveBundleIndex>` that is
+/// much cheaper when a bundle conflicts with many others (e.g. a
+/// function with many locals).
+#[derive(Clone, Debug, Default)]
+pub struct ConflictSet {
+    // Generation at which each bundle was last inserted. The value `0`
+    // means "never inserted"; `generation` is therefore always >= 1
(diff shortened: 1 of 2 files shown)


michael-weigelt:
@cfallin I was told you might want to have a look at this.

michael-weigelt:
This is a small and (maybe too) focused improvement in the direction that is also discussed here: https://github.com/bytecodealliance/regalloc2/issues/87

cfallin:
@michael-weigelt thanks for this. It's an interesting optimization. Could you build a Wasmtime engine (`wasmtime-bench-api` target) with this version of regalloc2 and with baseline, and test both with [Sightglass](https://github.com/bytecodealliance/sightglass) over the full suite?

Those speedups are frankly fantastic if real but I'm also more than a little skeptical about purported 2-3x speedups; I've done a lot of profiling of regalloc2 and, at least last time I did this (a few years ago now), didn't see major hotspots outside of the expected ones (e.g. the main allocation maps). That said, I welcome benchmark results that say otherwise if reliable!

Data-structure-wise, I was initially skeptical in reading the patch about the dense representation but I see that it's a reused set so we will only allocate one per function compilation -- so it's O(n) time and space cost overall, which is acceptable. It's still worse than the O(k) cost for k << n (very few conflicts), so I do want a careful evaluation over the suite.

Side-note: in BA we now have the [AI tools policy](https://github.com/bytecodealliance/governance/blob/main/AI_TOOL_POLICY.md), and while you are welcome to use tools such as Claude privately (many of us do for various purposes), you are fully and solely responsible for interactions and discussions here, so instead of "Claude said [thing you're unsure about]", please speak only about things you've personally understood/vetted and don't relay raw agent thoughts or output. In this case that vetting includes running the benchmarks, and carefully reasoning about the algorithmic tradeoff as humans ourselves. Thanks!

cfallin:
Also: I would want to see that Wasmtime+Cranelift pass all tests wth the modified regalloc2, just to make sure we're not missing something wrt changed output (which could also explain huge compile-time wins if e.g. we suddenly treat conflicts differently and allocate less optimally or whatever). The algorithm *looks* correct but let's verify.

michael-weigelt:
Thanks for the pointers, I'll run those benchmarks. I don't expect the speedups of the general cases to be this impressive, but if they are at least not worse, I am happy. 

> you are fully and solely responsible for interactions and discussions here, so instead of "Claude said [thing you're unsure about]", please speak only about things you've personally understood/vetted

Understood and agreed. The whole point of that sentence was: I want to vet this. For the record, because I am a new contributor here: The only AI contributions in this PR are the brainstorming on _how_ to address the problem and code. I would never relay slop to a human or let it talk/write prose for me.

src/ion/data_structures.rs
```diff
@@ -444,6 +444,53 @@ impl core::ops::IndexMut<VReg> for VRegs {
...
+    // after the first `clear`.
+    stamps: Vec<u32>,
+    generation: u32,
+}
+
+impl ConflictSet {
+    /// Empty the set. O(1) except on the rare generation wraparound.
+    #[inline]
+    pub fn clear(&mut self) {
+        self.generation = self.generation.wrapping_add(1);
```
michael-weigelt: We could instead use a u64 and ignore the risk of wrap around.

cfallin: Yes, definitely; with existing limits on function size it's actually not possible to have more than `u64::MAX` live bundles (nevermind the implied execution time that would require).

michael-weigelt:
> Wasmtime+Cranelift pass all tests wth the modified regalloc2

I ran them all (`cargo test` in the wasmtime/cranelift directory and also at the root) and observed no failures. 

I ran `all.suite` on Sightglass (excluding only splay.wasm because that seems to miss a gc flag somewhere), and the results are attached here. I ran it once with the default engine settings, and once with no optimizations (which is my use case). The results are very mixed. One way to summarize them: 

- With noopt, baseline is faster in 6 benchmarks and slower in 11 benchmarks than this PR.
- With opt, baseline is faster in 14 benchmarks and slower in 17 benchmarks than this PR. 

But it's quite variable by how much they differ. 

I am interested to hear your thoughts about this, @cfallin. I don't know what operations points are most important for this library, in particular idk if you care about the noopt case. Furthermore, the PR is motivated by a uncommon Wasm module, so I do realize I have the double disadvantage of a special operation point and a special input. All that is to say: I'd understand if you reject this PR. But in its favour at least I'd say it does not look clearly worse on average, and massively better in the very special case. 

[all_with_noopt.txt](https://github.com/user-attachments/files/28753443/all_with_noopt.txt)
[all_normal.txt](https://github.com/user-attachments/files/28753444/all_normal.txt)

src/ion/data_structures.rs
michael-weigelt: 14c22e7

cfallin:
Thanks, @michael-weigelt . I think the best way forward here is to be objective with the benefit of data: if the PR causes more compilation slowdowns than speedups, it's not worth it overall (for a general-purpose compiler). I guess I'd that this is not too surprising when the initial evaluation is a Wasm of a "special" shape -- undoubtedly one could specialize the compiler for certain corners of the parameter space and get better performance in those cases, but we unfortunately have to build a compiler that works well for most typical inputs. Thanks nevertheless for your efforts here!

michael-weigelt:
> if the PR causes more compilation slowdowns than speedups

It does not, it shuffles around performance. Some cases are up to 1.6x faster, others are up to 1.6x slower. There are more cases which are faster with the PR (in both the opt and noopt case). 

It's not clear to me if it's good enough overall, so please don't misunderstand this comment as pushing for the change. But I want to make sure we read the data the same way. Concretely: Is the problem that "we win some, we lose some", and that's not good enough, or is it that a particular _important_ subset of benchmarks suffer from this change?

cfallin:
Ah, I read "faster in 6 / slower in 11" and "faster in 14 / slower in 17" (no-opt and opt respectively) as "more compilation slowdowns than speedups". And, yes, in general the kinds of changes we'd take would ideally be mostly or entirely speedups -- a mixed result is not a clear win.

PR #262, opened 2026-07-06
Remove dead cache from `IndexSet`

fitzgen:
The `IndexSet::cache` field was never populated with a real value, and `cache.set()` was only ever called with the `(INVALID, 0)` sentinel, so the fast-path in `maybe_elem()` was unreachable and the invalidation stores in `elem()` and `maybe_elem_mut()` were pure-overhead on every `set`/`get`/`union_with`. Removing it additionally shrinks `IndexSet` by 16 bytes.

### DIFF
--- a/src/indexset.rs
+++ b/src/indexset.rs
@@ -6,7 +6,6 @@
 //! Index sets: sets of integers that represent indices into a space.
 
 use alloc::vec::Vec;
-use core::cell::Cell;
 
 use crate::FxHashMap;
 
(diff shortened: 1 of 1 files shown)


fitzgen:
(FWIW, I also experimented with fixing the cache so that it would actually get populated, but it wasn't a speed up on Sightglass.)

cfallin (approved):
Ah, the actual cache usage was removed a while ago -- thanks for catching this!

